PR #13, opened 2021-09-30
Add a pandas UDF predict function in MultiSeriesProphetModel

lu-wang-dl:


runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -193,15 +193,43 @@ def predict_timeseries(self, horizon: int = None) -> pd.DataFrame:
         ids = pd.DataFrame(self._model_json.keys(), columns=["ts_id"])
         return ids.groupby("ts_id").apply(lambda df: self._predict_impl(df, horizon)).reset_index()
 
-    def predict(self, context: mlflow.pyfunc.model.PythonModelContext, input: pd.DataFrame)-> pd.DataFrame:
+    @staticmethod
+    def get_reserved_cols():
+        reserved_names = [
```
panchalhp-db, line 198: Should we add this to a class variable?

panchalhp-db (approved):

runtime/databricks/automl_runtime/forecast/prophet/model.py
lu-wang-dl: This variable itself does not mean anything. The return of this function will give the actual reserved_names.

PR #14, opened 2021-11-01
Check whether the metric exist when logging

lu-wang-dl:


runtime/databricks/automl_runtime/forecast/prophet/forecast.py
```diff
@@ -164,7 +164,8 @@ def fit(self, df: pd.DataFrame) -> pd.DataFrame:
         results_pd = pd.DataFrame({"model_json": model_json}, index=[0])
         results_pd.reset_index(level=0, inplace=True)
         for metric in self.SUPPORTED_METRICS:
```
wenfeiy-db, line 166: Do we need to keep the SUPPORTED_METRICS the same as in `automl`?
See https://livegrep.dev.databricks.com/view/databricks/universe/automl/python/databricks/automl/forecast.py#L53

lu-wang-dl: Good catch. Make them the same.

wenfeiy-db: Oh just made a closer look. Then do we also need to change the `result_columns` in notebooks?
In `base_prophet.jinja` [here](https://livegrep.dev.databricks.com/view/databricks/universe/automl/python/databricks/automl/sections/templates/forecast/base_prophet.jinja#L65) we indexes the `results_pd` with `result_columns`, and the ``result_columns` include both mape and coverage (see [univaraite](https://livegrep.dev.databricks.com/view/databricks/universe/automl/python/databricks/automl/sections/templates/forecast/univariate_prophet.jinja#L12), [multivaraite](https://livegrep.dev.databricks.com/view/databricks/universe/automl/python/databricks/automl/sections/templates/forecast/multivariate_prophet.jinja#L26)).
Or maybe we can actually keep the `SUPPORTED_METRICS` unchanged, and use `np.nan` if a metric is not in `metrics.keys()`. I think this is how we dealt with it previously (https://github.com/databricks/universe/pull/113179/files).

wenfeiy-db (commented):
☝️ follow-up comment for `SUPPORTED_METRICS`

wenfeiy-db (approved):
lgtm!

PR #17, opened 2021-11-30
Change the way to setup random state in hyperopt

lu-wang-dl:
The new hyperopt changed the way to set the random seed. This PR is to update the way to call hyperopt.


panchalhp-db (approved):
LGTM

PR #18, opened 2021-12-14
Update predict API to match MLflow API

lu-wang-dl:


runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -61,19 +64,35 @@ class ProphetModel(mlflow.pyfunc.PythonModel):
...
+        :param target_col: the column name of the predict target column
         """
         self._model_json = model_json
         self._horizon = horizon
         self._frequency = frequency
+        self._time_col = time_col
+        self._target_col = target_col
         super().__init__()
 
+    def _validate_id_cols(self, df: pd.DataFrame, required_cols: List[str]):
```
wenfeiy-db, line 85: Shall we call this `_validate_cols` instead?

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -117,35 +136,45 @@ def predict_timeseries(self, horizon: int = None) -> pd.DataFrame:
         """
```
wenfeiy-db, line 136: I felt the name `predict_timeseries` is not clear enough. It's not obvious how it's different than the `predict` method. Maybe we can call it `forecast` and clarify in the docstring that this method makes multiple-step forecasting and `horizon` is the number of time steps.

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -61,19 +64,35 @@ class ProphetModel(mlflow.pyfunc.PythonModel):
     """
```
wenfeiy-db, line 64: Do we need to update the version of `databricks-automl-runtime` in `PROPHET_CONDA_ENV`?

runtime/databricks/automl_runtime/forecast/prophet/model.py
lu-wang-dl: The original design for `predict_timeseries` is basically the same with `predict` but only take the `horizon`. Right now the major difference is that `predict_timeseries` keeps the extra columns from prophet. We can keep this now for backward compatibility.

wenfeiy-db: i just realized that `predict_timeseries` provides predictions for the given horizon as well as in-sample data. Shall we add another method `forecast` that only does out-of-sample forecasting?

lu-wang-dl: Let's do this as a follow up.

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -117,35 +136,45 @@ def predict_timeseries(self, horizon: int = None) -> pd.DataFrame:
...
         Predict API from mlflow.pyfunc.PythonModel
         :param context: A :class:`~PythonModelContext` instance containing artifacts that the model
                         can use to perform inference.
-        :param input: Input dataframe
+        :param input_df: Input dataframe
         :return: A pd.DataFrame with the forecast components.
         """
-        return self._predict_impl()
+        self._validate_cols(input_df, [self._time_col])
+        input_df.rename(columns={self._time_col: "ds"}, inplace=True)
```
yxiong, line 148: [Discussion] Why do we want to modify the `input_df`? I don't think it's consistent with the API other problem types. Users typically do `y = model.predict(X)`, expect `X` to be unchanged and `y` to be a new Series or DataFrame. I think this is the behavior of prophet's own `predict` function as well [code](https://github.com/facebook/prophet/blob/ba9a5a2c6e2400206017a5ddfd71f5042da9f65b/python/prophet/forecaster.py#L1205) (note `df.copy()`). Shall we follow that convention?

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -61,19 +64,35 @@ class ProphetModel(mlflow.pyfunc.PythonModel):
     """
     Prophet mlflow model wrapper for univariate forecasting.
     """
-    def __init__(self, model_json: Union[Dict[str, str], str], horizon: int, frequency: str) -> None:
+    def __init__(self, model_json: Union[Dict[str, str], str], horizon: int, frequency: str,
+                 time_col: str, target_col: str) -> None:
```
yxiong, line 68: I don't think `target_col` should be part of the API. It's more natural to do `y = model.predict(X)`, with the output returned rather than written back to `X` in place.

runtime/databricks/automl_runtime/forecast/prophet/model.py
lu-wang-dl: The main issue is that prophet requires the input column name be ds (timestamp) and y(target).

runtime/databricks/automl_runtime/forecast/prophet/model.py
lu-wang-dl: The issue is that prophet model does not save the target column name and time column name. We have to do it in order to make the behavior the same with classification and regression.

runtime/databricks/automl_runtime/forecast/prophet/model.py
wenfeiy-db: Could we make a copy of input_df instead of directly modifying it? From the linked code by Ying, that's what prophet does.

yxiong: I understand we need to rename the input column to `ds`, but we do not need to specify target column name at inference time, right? What I have in mind is

```
def predict(self, input_df):
  ds_df = pd.DataFrame({"ds": input_df[self._time_col]))  # Make a copy
  predict_df = self.model().predict(ds_df)
  return predict_df["yhat"]
```

Note that we return a `pd.Series`, not a `pd.DataFrame`. This is consistent with classification/regression, where user can either do `y = model.predict(X)` or `X["result"] = model.predict(X)`. What do you think?

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -117,35 +136,45 @@ def predict_timeseries(self, horizon: int = None) -> pd.DataFrame:
         """
         return self._predict_impl(horizon)
 
-    def predict(self, context: mlflow.pyfunc.model.PythonModelContext, input: pd.DataFrame) -> pd.DataFrame:
+    def predict(self, context: mlflow.pyfunc.model.PythonModelContext, input_df: pd.DataFrame) -> pd.DataFrame:
```
yxiong, line 139: It doesn't seem we need `context` in this function?

lu-wang-dl: The context is required for mlflow.

runtime/databricks/automl_runtime/forecast/prophet/model.py
lu-wang-dl: It is fine with univariate forecast. But for multi-series forecasting, we have to return a dataframe with id columns. In this case, we need to return the predicted value with column name. Any better suggestion on this?

yxiong: I think we can also return a `pd.Series` in multi-series forecasting, where the index of the returned series maps to the index of input dataframe one-by-one. If the user wants to add result back to their input, they can always do `X["result"] = model.predict(X)`, but by default, they get a result without the input modified `y = model.predict(X)`.

We can also check with ml-api@databricks.com for suggestions. +@mengxr .

runtime/databricks/automl_runtime/forecast/prophet/model.py
yxiong: Can you tell me more about this context? The `mlflow.pyfunc.PyFuncModel` API does not seem to contain that part: https://github.com/mlflow/mlflow/blob/044ad9e545f4ee7ac23596a9a890bf5b30aac490/mlflow/pyfunc/__init__.py#L589

lu-wang-dl: It is trying to implement the function: https://github.com/mlflow/mlflow/blob/0fa849ad75e5733bf76cc14a4455657c5c32f107/mlflow/pyfunc/model.py#L86

lu-wang-dl: You may check the moflow doc for some examples: https://github.com/mlflow/mlflow/blob/19b3ad1009b668827a56be645caa5fd0280cab19/docs/source/models.rst#custom-python-models

yxiong: I see. Thanks for the pointer! One small nitpick: we should keep the parameter name same to the base class, i.e. `model_input` instead of `input_df` -- parameter name is part of the python API.

runtime/databricks/automl_runtime/forecast/prophet/model.py
yxiong: Can we get rid of the `target_col` parameter now? :)

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -117,35 +136,44 @@ def predict_timeseries(self, horizon: int = None) -> pd.DataFrame:
...
         Predict API from mlflow.pyfunc.PythonModel
         :param context: A :class:`~PythonModelContext` instance containing artifacts that the model
                         can use to perform inference.
-        :param input: Input dataframe
+        :param model_input: Input dataframe
         :return: A pd.DataFrame with the forecast components.
         """
-        return self._predict_impl()
+        self._validate_cols(model_input, [self._time_col])
+        model_input.rename(columns={self._time_col: "ds"}, inplace=True)
```
yxiong, line 148: Instead of rename, let's do `X = pd.DataFrame{"ds": model_input[self._time_col]}`, so that the model input is not modified?

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -225,15 +253,21 @@ def model_predict(self, df: pd.DataFrame, horizon: int = None) -> pd.DataFrame:
...
         Predict API from mlflow.pyfunc.PythonModel
         :param context: A :class:`~PythonModelContext` instance containing artifacts that the model
                         can use to perform inference.
-        :param input_df: Input dataframe
+        :param model_input: Input dataframe
         :return: A pd.DataFrame with the forecast components.
         """
-        return self._predict_impl(input_df)
+        self._validate_cols(model_input, self._id_cols + [self._time_col])
+        model_input["ts_id"] = model_input[self._id_cols].agg('-'.join, axis=1)
```
yxiong, line 265: ditto, can we create a separate copy `X` instead of directly modifying the `model_input`?

runtime/tests/automl_runtime/prophet/model_test.py
```diff
@@ -80,9 +82,52 @@ def test_model_save_and_load_multi_series(self):
...
 
-        # Check model_predict function
-        prophet_model._model_impl.python_model.model_predict(ids)
+        # Check predict API
+        num_rows = 2
+        test_df = pd.concat([
+            pd.to_datetime(pd.Series(range(num_rows), name="time").apply(lambda i: f"2020-11-{3*i+1}")),
+            pd.Series(range(num_rows), name="id").apply(lambda i: f"{i%2+1}")
+        ], axis=1)
+        forecast_y = prophet_model.predict(test_df)
```
yxiong, line 95: Nit: if there are only a couple of rows, I recommend just write it out explicitly, something like

```python
test_df = pd.DataFrame({
  "time": [pd.to_datetime("2020-11-01"), pd.to_datetime("2021-11-04")],
  "id": [1,2],
})```

BTW, if the `time` column input is string but not timestamp, would the API work?

File: runtime/databricks/automl_runtime/forecast/prophet/model.py (hunk 7 above)
Comment (lu-wang-dl) [reply runtime/databricks/automl_runtime/forecast/prophet/model.py:148]: Since we didn't return model_input, the user will not see the changes, right?

Comment (yxiong) [reply runtime/databricks/automl_runtime/forecast/prophet/model.py:148]: No, python function doesn't do pass-parameter-by-value. Here's an example

```
def func(df):
  df.rename(columns:{"foo", "bar"}, inplace=True)

df = pd.DataFrame({"foo": [1,2, 3,4])
df

```

File: runtime/tests/automl_runtime/prophet/model_test.py (hunk 9 above)
Comment (lu-wang-dl) [reply runtime/tests/automl_runtime/prophet/model_test.py:94]: Not for now. We have a followup ticket to fix this.

File: runtime/tests/automl_runtime/prophet/model_test.py [hunk 10]
```diff
@@ -80,9 +82,51 @@ def test_model_save_and_load_multi_series(self):
...
+        np.testing.assert_array_almost_equal(np.array(forecast_y),
+                                             np.array([10.333333, 11.333333]))
+
+    def test_validate_predict_cols(self):
+        prophet_model = ProphetModel(self.model_json, 1, "d", "time")
+        test_df = pd.concat([
+            pd.to_datetime(pd.Series(range(2), name="ds").apply(lambda i: f"2020-11-{3*i+1}")),
+            pd.Series(range(2), name="id").apply(lambda i: f"{i%2+1}")
+        ], axis=1)
```
yxiong, line 103: Let's also rewrite this `test_df` and the one below for better readability.

runtime/tests/automl_runtime/prophet/model_test.py
yxiong: Maybe add a assert to make sure `test_df` is not modified? Just to guard against future mistakes :)

yxiong (approved):
Thanks for addressing my comments. LGTM!

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -225,15 +249,22 @@ def model_predict(self, df: pd.DataFrame, horizon: int = None) -> pd.DataFrame:
...
-        :param input_df: Input dataframe
+        :param model_input: Input dataframe
         :return: A pd.DataFrame with the forecast components.
         """
-        return self._predict_impl(input_df)
+        self._validate_cols(model_input, self._id_cols + [self._time_col])
+        test_df = model_input.copy()
+        test_df["ts_id"] = test_df[self._id_cols].agg('-'.join, axis=1)
+        test_df.rename(columns={self._time_col: "ds"}, inplace=True)
+        group_cols = ["ts_id"] + self._id_cols
```
wenfeiy-db, line 264: Why do we need to group by on both ts_id and the id_cols? I thought it'll be the same as just grouping by ts_id?

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -225,15 +249,22 @@ def model_predict(self, df: pd.DataFrame, horizon: int = None) -> pd.DataFrame:
         result_df["ts_id"] = str(df["ts_id"].iloc[0])
         return result_df[return_cols]
 
-    def predict(self, context: mlflow.pyfunc.model.PythonModelContext, input_df: pd.DataFrame) -> pd.DataFrame:
+    def predict(self, context: mlflow.pyfunc.model.PythonModelContext, model_input: pd.DataFrame) -> pd.DataFrame:
```
wenfeiy-db, line 252: Just to make sure. Is the conclusion that we'll return a pd.Series of yhat for the predict API instead of a pd.DataFrame?
In this case, shall we change the type hints of the predict methods? For both MultiSeriesProphetModel and ProphetModel.

wenfeiy-db (approved):

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -225,15 +249,22 @@ def model_predict(self, df: pd.DataFrame, horizon: int = None) -> pd.DataFrame:
...
+        :param model_input: Input dataframe
         :return: A pd.DataFrame with the forecast components.
         """
-        return self._predict_impl(input_df)
+        self._validate_cols(model_input, self._id_cols + [self._time_col])
+        test_df = model_input.copy()
+        test_df["ts_id"] = test_df[self._id_cols].agg('-'.join, axis=1)
+        test_df.rename(columns={self._time_col: "ds"}, inplace=True)
+        group_cols = ["ts_id"] + self._id_cols
+        predict_df = test_df.groupby(group_cols).apply(lambda df: self.model(df.name[0]).predict(df)).reset_index()
```
wenfeiy-db, line 265: Also, shall we catch the exception caused by unseen ts_id to show meaningful error message?

runtime/databricks/automl_runtime/forecast/prophet/model.py
lu-wang-dl: It is mainly used for joining with the orignal dataframe.

runtime/databricks/automl_runtime/forecast/prophet/model.py
lu-wang-dl: It is checked with _validate_cols in line 260.

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -225,15 +249,22 @@ def model_predict(self, df: pd.DataFrame, horizon: int = None) -> pd.DataFrame:
...
+        return_df = test_df.merge(predict_df, how="left", on=self._id_cols)
```
wenfeiy-db, line 266: wait, why do we merge only on id_cols? I think we should also merge on time_col. For example if we have a dataframe with `time_col = [pd.TimeSeries(2018, 10, 1), pd.TimeSeries(2018, 10, 2), pd.TimeSeries(2018, 10, 1), pd.TimeSeries(2018, 10, 2)]` and `id_col = [1, 1, 2, 2]`, merging only on id_cols will cause problems.

yxiong: I believe merging `id_cols` is the right thing, as it will put `[pd.Timestamp(2018, 10, 1), pd.Timestamp(2018, 10, 2)]` into a `df` and pass it into the lambda. We should add a test case to verify the behavior though.

wenfeiy-db: Could you explain? Say we have
```
test_pd = pd.DataFrame({"ds": [pd.Timestamp(2018, 10, 1), pd.Timestamp(2018, 10, 2), pd.Timestamp(2018, 10, 1), pd.Timestamp(2018, 10, 2)], "id": [1, 1, 2, 2]})
predict_df = pd.DataFrame({"ds": [pd.Timestamp(2018, 10, 1), pd.Timestamp(2018, 10, 2), pd.Timestamp(2018, 10, 1), pd.Timestamp(2018, 10, 2)], "id": [1, 1, 2, 2], "yhat": [1, 2, 3, 4]})
```
Then running `test_df.merge(predict_df, how="left", on=["id",])` returns a dataframe with 8 rows.
Did I get something wrong?

yxiong: Ah, you're right. This merge in L266 should also include the `ds` column. I though we were talking about the `groupby` in L265. Either way, having a unit test would be good @lu-wang-dl :)

lu-wang-dl: Good point. I will update it.

wenfeiy-db (approved):
LGTM!

PR #20, opened 2022-01-06
Add model wrapper, training module, and diagnostics functions for pmdarima

wenfeiy-db:
* As titled
* Also adds imputation for missing time steps in the training module

Manual tests
* Univariate: https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/3959285228605954/command/3959285228605983
* Multivariate: https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/3959285228602434/command/3959285228602456


runtime/databricks/automl_runtime/forecast/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+    "hours": "H",
+    "hour": "H",
+    "hr": "H",
+    "h": "H",
+    "H": "H",
+    "m": "min",
+    "minute": "min",
+    "min": "min",
+    "minutes": "min",
+    "T": "T",
```
yxiong, line 32: Just curious, what's 'T'?

runtime/databricks/automl_runtime/forecast/pmdarima/diagnostics.py
```diff
@@ -0,0 +1,111 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import pandas as pd
+import numpy as np
+
+
+def generate_cutoffs(df: pd.DataFrame, horizon: int, unit: str, num_folds: int):
```
yxiong, line 21: Is this specific to pmdarima? Can we move it one level up as as a forecast util and share with prophet?

ditto for the functions below.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -0,0 +1,323 @@
...
+        :param end_ds: the end time of training data
+        :param time_col: the column name of the time column
+        """
+        self._pickled_model = pickled_model
+        self._horizon = horizon
+        self._frequency = frequency
+        self._start_ds = start_ds
+        self._end_ds = end_ds
+        self._time_col = time_col
+        super().__init__()
```
yxiong, line 48: Nit: prefer to call `super().__init__()` at the beginning, unless you need some do some preprocessing and pass the result to `super().__init__()`.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -0,0 +1,323 @@
...
+        """
+        Loads artifacts from the specified PythonModelContext that can be used by
+        PythonModel.predict when evaluating inputs. When loading an MLflow model with
+        load_pyfunc, this method is called as soon as the PythonModel is
+        constructed.
+        :param context: A PythonModelContext instance containing artifacts that the model
+                        can use to perform inference.
+        """
+        from pmdarima.arima import ARIMA  # noqa: F401
+        return
```
yxiong, line 60: Nit: the `return` is unnecessary.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -0,0 +1,323 @@
...
+        :return: A pd.DataFrame with the forecasts and confidence intervals for given horizon and history data.
+        """
+        horizon = horizon or self._horizon
+        future_pd = self._forecast(horizon)
+        in_sample_pd = self._predict_in_sample()
+        return pd.concat([in_sample_pd, future_pd]).reset_index(drop=True)
+
+    def predict(self, context: mlflow.pyfunc.model.PythonModelContext, model_input: pd.DataFrame) -> pd.Series:
+        """
+        Predict API from mlflow.pyfunc.PythonModel. Returns the prediction values for given timestamps in the input
```
yxiong, line 83: Nit: prefer to have a one-line summary first, followed by an empty line, followed by detailed documentation. https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

ditto for other docstrings.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -0,0 +1,323 @@
...
+        if not required_cols_set.issubset(df_cols):
+            raise MlflowException(
+                message=(
+                    f"Input data columns '{list(df_cols)}' do not contain the required columns '{required_cols}'"
+                ),
+                error_code=INVALID_PARAMETER_VALUE,
+            )
+
+
+class MultiSeriesArimaModel(mlflow.pyfunc.PythonModel):
```
yxiong, line 164: Looks like there are quite some code duplication with uni-variate `ArimaModel`. Is there a way we can reduce that? e.g. create a list of `ArimaModel`s inside `MultiSeriesArimaModel` and use a for loop to call into the method inside.

runtime/databricks/automl_runtime/forecast/pmdarima/training.py
```diff
@@ -0,0 +1,112 @@
...
+        self._seasonal_periods = seasonal_periods
+        self._num_folds = num_folds
+
+    def fit(self, df: pd.DataFrame) -> pd.DataFrame:
+        """
+        Fit the ARIMA model with tuning of seasonal period m and with pmdarima.auto_arima.
+        :param df: A pd.DataFrame containing the history data. Must have columns ds and y.
+        :return: A pd.DataFrame with the best model (pickled) and its metrics from cross validation.
+        """
+        import pandas as pd
```
yxiong, line 52: pandas is already imported.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
lu-wang-dl: +1

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -0,0 +1,323 @@
...
+        """
+        from pmdarima.arima import ARIMA  # noqa: F401
+        return
+
+    def model(self) -> pmdarima.arima.ARIMA:
+        """
+        Deserialize the ARIMA model by pickle.
+        :return: ARIMA model
+        """
+        import pickle
```
lu-wang-dl, line 67: Could we load the module at the beginning of the file?

runtime/databricks/automl_runtime/forecast/pmdarima/training.py
```diff
@@ -0,0 +1,112 @@
...
+        import pickle
```
lu-wang-dl, line 53: We don't need to import these in the function, right?

runtime/requirements.txt
```diff
@@ -3,10 +3,12 @@
 
 holidays
 hyperopt
+koalas
```
lu-wang-dl, line 6: Why do we need koalas?

lu-wang-dl: This will include spark as dependency. And for new spark, koalas is already in spark. We don't want to include this.

lu-wang-dl (changes requested):

runtime/databricks/automl_runtime/forecast/__init__.py
wenfeiy-db: "T" is the same as "min". Probably we should keep it consistent to map to "min".

runtime/databricks/automl_runtime/forecast/pmdarima/diagnostics.py
wenfeiy-db: Yes, the two functions used for pmdarima and prophet are similar. I didn't make them a common one because I still need to modify the pmdarima one when adding the tuning of m later (which is likely not going to be in 10.3), so I'd like to keep the existing prophet one as is for now to make sure I won't break things for 10.3.
So I'd prefer keeping the functions for pmdarim and prophet separate in this PR. WDYT?

runtime/databricks/automl_runtime/forecast/pmdarima/training.py
wenfeiy-db: I felt I did this previously to make the multiseries forecasitng work. The prophet estimator is also doing something similar (https://github.com/databricks/automl/blob/468206f3161546beeeb5bff1cc1c46d69accc41d/runtime/databricks/automl_runtime/forecast/prophet/forecast.py#L124). Let me double-check though.

runtime/requirements.txt
wenfeiy-db: I added it because I saw it's in the `environment.txt`. Is there a reason why we want it in `environemnt.txt` but not in `requriements.txt`?

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
wenfeiy-db: I was following the style used in the propeht model wrapper ([here](https://github.com/databricks/automl/blob/468206f3161546beeeb5bff1cc1c46d69accc41d/runtime/databricks/automl_runtime/forecast/prophet/model.py#L108)). I probably should have asked earlier whether it's a requirement for the `mlflow.pyfunc.PythonModel`. 
Let me fix both.

runtime/databricks/automl_runtime/forecast/__init__.py
yxiong: Yes, I think it should map to `min`. Otherwise when you `_fill_missing_time_steps` and do `resample` with `T`, would it work?

runtime/databricks/automl_runtime/forecast/pmdarima/diagnostics.py
yxiong: I feel it will be a maintenance burden in the long run if we kept them separate and have the logics diverge. IIRC, this is a generic utility for forecasting in general, and thus we should avoid the temptation to customize for different framework. We should make such utility functions universally well-defined (in terms of the forecasting problem, not in terms of individual solver), well-documented and well-tested.

runtime/databricks/automl_runtime/forecast/__init__.py
wenfeiy-db: Replaced. BTW T does work with the pandas resample function.

runtime/databricks/automl_runtime/forecast/pmdarima/diagnostics.py
wenfeiy-db: I agree that we definitely want to use common `generate_cutoffs` function (and other similar utility functions), which will also make it easier for us to extend with other forecasting models later. I have a backlog ticket (https://databricks.atlassian.net/browse/ML-19542) for cleaning up code, which I planned to do after adding the tuning for m (https://databricks.atlassian.net/browse/ML-19328). Does it sound good to make the changes with that ticket?

runtime/databricks/automl_runtime/forecast/pmdarima/training.py
wenfeiy-db: It doesn't have to be inside the method. Checked and fixed.

runtime/requirements.txt
wenfeiy-db: As per offline discussion, `environment.txt` is only used by github workflow for test set up. But we don't want to add koalas for the project dependencies with `requirements.txt`.

runtime/databricks/automl_runtime/forecast/pmdarima/diagnostics.py
yxiong: I personally think we should unify early and then decide what to fork, as it'll be easier than the other way around. But I agree we don't have to do it in this PR.

yxiong (approved):
LGTM! I still think we should spend some effort deduplicate common util functions, but we can leave it to a future PR :)

lu-wang-dl:
Let's make a followup to improve the test coverage:
```
databricks/automl_runtime/forecast/pmdarima/diagnostics.py          48      4    92%
databricks/automl_runtime/forecast/pmdarima/model.py               117      1    99%
databricks/automl_runtime/forecast/pmdarima/training.py             48      1    98%
databricks/automl_runtime/forecast/pmdarima/utils.py                35     22    37%
```

runtime/databricks/automl_runtime/forecast/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+    "m": "min",
+    "minute": "min",
+    "min": "min",
+    "minutes": "min",
+    "T": "min",
+    "S": "S",
+    "seconds": "S",
+    "sec": "S",
+    "second": "S"
+}
```
lu-wang-dl, line 37: nit: extra line at the end.

lu-wang-dl (approved):

lu-wang-dl:
> LGTM! I still think we should spend some effort deduplicate common util functions, but we can leave it to a future PR :)

Let's make sure we have this as followup too.

wenfeiy-db:
> > LGTM! I still think we should spend some effort deduplicate common util functions, but we can leave it to a future PR :)
> 
> Let's make sure we have this as followup too.

> Let's make a followup to improve the test coverage:
> 
> ```
> databricks/automl_runtime/forecast/pmdarima/diagnostics.py          48      4    92%
> databricks/automl_runtime/forecast/pmdarima/model.py               117      1    99%
> databricks/automl_runtime/forecast/pmdarima/training.py             48      1    98%
> databricks/automl_runtime/forecast/pmdarima/utils.py                35     22    37%
> ```

Added description for both in this clean up ticket https://databricks.atlassian.net/browse/ML-19542

Issue #19, opened 2022-01-04
Prophet AutoML never creates model with best parameters from Hyperopt


rbloehm:
There is a bug using the ProphetHyperParams enum in the AutoML implementation for forecasting with prophet.
**The metrics report the potentially very good best result from Hyperopt, but the stored model itself, the evaluation plots and the forecast results are completely unrelated and probably much worse.**

This leads to the following situation which happens every time someone uses Databricks AutoML for forecasting:
* Run Hyperopt, find a best set of parameters (best_result), for example: 
![image](https://user-images.githubusercontent.com/10888961/148073451-a6da3fd4-c83f-47ff-a001-c21db4387fd9.png)
note: seasonality_mode=1 is multiplicative.
* The metrics reported in Databricks AutoML & mlflow are the ones from this best_result. So far so good.
* The model that is then trained on the full dataset, evaluated in the Training notebook (plots etc.) and stored as artifact **has nothing to do with those selected hyperparameters**: 
![image](https://user-images.githubusercontent.com/10888961/148074030-0a98e8f6-2c03-49b6-9cf6-c2959a5b54ec.png)

**Potential reason / bug:**
https://github.com/databricks/automl/blob/242bf1afbff43a0f6e94e29bee7f01552489a004/runtime/databricks/automl_runtime/forecast/prophet/forecast.py#L150-L153
![image](https://user-images.githubusercontent.com/10888961/148074387-1be3ae12-f4b3-48cb-a046-278846c0ecfe.png)

Databricks Runtime 10.2 ML
databricks-automl-runtime==0.2.4

wenfeiy-db:
@rbloehm Thanks for reporting this! It will be fixed in new release. We will also patch previous releases which were affected.

rbloehm:
@wenfeiy-db I installed the new version 0.2.4.1 in one of our clusters and can confirm that the bug has been fixed. Thanks!

PR #22, opened 2022-01-10
Fix param enum for prophet

wenfeiy-db:
Fixes issue reported by https://github.com/databricks/automl/issues/19

## Test
Manual checking [notebook](https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/884158634248354/command/884158634248841). Confirmed the best model is not using default params

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/prophet/forecast.py
+++ b/runtime/databricks/automl_runtime/forecast/prophet/forecast.py
@@ -146,6 +146,8 @@ class ProphetHyperoptEstimator(ABC):
             timeout=self._timeout,
             rstate=self._random_state)
 
+        print("best_result:", best_result)
+
         # Retrain the model with all history data.
         model = Prophet(changepoint_prior_scale=best_result.get(ProphetHyperParams.CHANGEPOINT_PRIOR_SCALE.value, 0.05),
                         seasonality_prior_scale=best_result.get(ProphetHyperParams.SEASONALITY_PRIOR_SCALE.value, 10.0),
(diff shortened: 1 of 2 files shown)


panchalhp-db (approved):
Is it possible to add a unit test to ensure this doesn't break in the future?

wenfeiy-db:
> Is it possible to add a unit test to ensure this doesn't break in the future?

I couldn't think of an easy way. Does it make sense that we train a model and check the best params are not all equal to default values? I felt this is not deterministic.

rbloehm:
You could do something here in the existing test: https://github.com/wenfeiy-db/automl/blob/29291477b0bb14f9b463933ad2601fa2547464b3/runtime/tests/automl_runtime/prophet/forecast_test.py#L34-L53

Change the search space in way such that the default parameter (0.05, respectively ln(0.05) ~ -3) is not included
ln(0.1) < -2.3 < -0.7 < ln(0.5)
```python
self.search_space = {"changepoint_prior_scale": hp.loguniform("changepoint_prior_scale", -2.3, -0.7)}
```

Check whether the finally used model_json's parameter is between 0.1 and 0.5 (and thus NOT the default, which is 0.05). (Code not tested, adapt accordingly):
```python
import json
model_dict = json.loads(results["model_json"].to_list()[0])
self.assertGreaterEqual(model_dict["changepoint_prior_scale"], 0.1)
self.assertLessEqual(model_dict["changepoint_prior_scale"], 0.5)
```

PR #25, opened 2022-01-12
Logging mlflow prophet model with signature

lu-wang-dl:
- Add a function `infer_signature` in the model class
- Update `mlflow_prophet_log_model` to log with signature

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/prophet/model.py
+++ b/runtime/databricks/automl_runtime/forecast/prophet/model.py
@@ -52,7 +52,8 @@ PROPHET_CONDA_ENV = {
         {
             "pip": [
                 f"prophet=={prophet.__version__}",
-                f"cloudpickle=={cloudpickle.__version__}"
+                f"cloudpickle=={cloudpickle.__version__}",
+                f"databricks-automl-runtime==0.2.5",
             ]
         }
     ],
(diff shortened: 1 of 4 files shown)


runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -51,8 +52,7 @@
         {
             "pip": [
                 f"prophet=={prophet.__version__}",
-                f"cloudpickle=={cloudpickle.__version__}",
-                f"databricks-automl-runtime==0.2.5"
+                f"cloudpickle=={cloudpickle.__version__}"
```
wenfeiy-db, line 55: Just for my understanding. What exactly is needed to be included in this conda_env used by `mlflow.pyfunc.log_model`?

runtime/tests/automl_runtime/prophet/model_test.py
```diff
@@ -112,26 +114,30 @@ def test_validate_predict_cols(self):
         run_id = run.info.run_id
         prophet_model = mlflow.pyfunc.load_model(f"runs:/{run_id}/model")
 
-        with pytest.raises(MlflowException, match="Input data columns") as e:
+        with pytest.raises(MlflowException, match="Model is missing inputs") as e:
```
wenfeiy-db, line 117: Is this error changed because we log the model with schema and mlflow find that the input dataframe doesn't have the "ds" column?

wenfeiy-db (approved):

runtime/tests/automl_runtime/prophet/model_test.py
lu-wang-dl: New error message is from mlflow

runtime/databricks/automl_runtime/forecast/prophet/model.py
lu-wang-dl: This environment is for model serving

PR #26, opened 2022-01-26
Fix ARIMA missing time step filling & forecasting for weekly dataset

wenfeiy-db:
## Changes
Fix a bug in the ARIMA model training for weekly datasets. Previously if the weekly dates is not on a Sunday, then the index will be resampled incorrectly because `df.resample(rule='W')` by default resample dates to Sundays, and similarly for `pd.date_range`.
Also add validation in `ArimaEstimator` for the training data - the timestamps in training data should have consistent frequency and the frequency should be the same as the one specified at the estimator creation.

## Tests
* Adds unit tests with weekly dataset
* Manual tested that previously failed AutoML ARIMA [trial](https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#job/410846556112865/run/1) will succeeds with updates in this PR (i.e. run `%pip install --force-reinstall --upgrade "git+https://github.com/wenfeiy-db/automl.git@fix-weekly-main#subdirectory=runtime"` at the beginning of the notebook)
  * notebook: https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/2151899606409450/command/2151899606409489


runtime/databricks/automl_runtime/forecast/pmdarima/training.py
```diff
@@ -106,4 +110,18 @@ def _fit_predict(df: pd.DataFrame, cutoffs: List[pd.Timestamp], seasonal_period:
...
-        return df.set_index("ds").resample(rule=OFFSET_ALIAS_MAP[frequency]).pad().reset_index()
+        df_filled = df.set_index("ds").resample(rule=OFFSET_ALIAS_MAP[frequency]).pad().reset_index()
+        start_ds, modified_start_ds = df["ds"].min(), df_filled["ds"].min()
+        if start_ds != modified_start_ds:
+            offset = modified_start_ds - start_ds
+            df_filled["ds"] = df_filled["ds"] - offset
+        return df_filled
+
+    @staticmethod
+    def _validate_ds_freq(df: pd.DataFrame, frequency: str):
```
wenfeiy-db, line 121: The check in `ArimaModel._predict_impl` and here should ideally use a common function. Will address it in the clean up ticket.
Same for the `start_ds` and `modified_start_ds` check in `ArimaEstimator._fill_missing_time_steps` and `AbstractArimaModel._get_ds`.

runtime/databricks/automl_runtime/errors.py
```diff
@@ -0,0 +1,15 @@
...
+    """
+    Base exception class for AutoML Runtime errors.
+    """
+
+    def __init__(self, message="", *args):
+        self.message = message
+        super().__init__(message, *args)
+
+
+class InvalidArgumentError(AutoMLRuntimeError):
```
yxiong, line 11: If we don't have lots special logic for the exception, can we just raise `ValueError`? https://stackoverflow.com/a/256260/4195568

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -55,6 +55,15 @@ def _validate_cols(df: pd.DataFrame, required_cols: List[str]):
                 error_code=INVALID_PARAMETER_VALUE,
             )
 
+    @staticmethod
+    def _get_ds(start_ds: pd.Timestamp, periods: int, frequency: str):
```
yxiong, line 59: Let's add a docstring for this function. Also, shall we call it `_get_ds_indices`?

yxiong (approved):

runtime/databricks/automl_runtime/errors.py
wenfeiy-db: Yes that makes sense!

PR #31, opened 2022-02-08
Update model logging logic for forecast models

lu-wang-dl:
Supoort sample input for pmarima model. Cleanup the model logging logic.

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/model.py
+++ b/runtime/databricks/automl_runtime/forecast/model.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-import pickle
 from abc import ABC, abstractmethod
 from typing import List
 
--- a/runtime/databricks/automl_runtime/forecast/model.py
+++ b/runtime/databricks/automl_runtime/forecast/model.py
@@ -28,8 +28,8 @@ class ForecastModel(ABC, mlflow.pyfunc.PythonModel):
     def __init__(self):
         super().__init__()
 
-    @abstractmethod
     @property
+    @abstractmethod
     def model_env(self):
         pass
 
--- a/runtime/databricks/automl_runtime/forecast/model.py
+++ b/runtime/databricks/automl_runtime/forecast/model.py
@@ -1,0 +1,67 @@
+#
+# Copyright (C) 2022 Databricks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+import pickle
+from abc import ABC, abstractmethod
+from typing import List
+
+import pandas as pd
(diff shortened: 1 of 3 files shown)


runtime/databricks/automl_runtime/forecast/model.py
```diff
@@ -0,0 +1,66 @@
...
+        if not required_cols_set.issubset(df_cols):
+            raise MlflowException(
+                message=(
+                    f"Input data columns '{list(df_cols)}' do not contain the required columns '{required_cols}'"
+                ),
+                error_code=INVALID_PARAMETER_VALUE,
+            )
+
+    def infer_signature(self, sample_input: pd.DataFrame = None) -> ModelSignature:
+        signature = infer_signature(sample_input, self.predict(context=None, model_input=sample_input))
```
wenfeiy-db, line 49: Why do we use `predict` instead of `predict_timeseries` here? I think calling `predict` with `model_input=None` will fail, at least for the `ArimaModel`. Could we add unit tests for `infer_signature` to verify if it work as expected?
Also I did see the example notebook where the model works for inference with pyspark dataframe. Does it mean that it actually works without the signature?

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -296,9 +290,11 @@ def _predict_single_id(self, df: pd.DataFrame) -> pd.DataFrame:
         return df
 
 
-def mlflow_arima_log_model(arima_model: Union[ArimaModel, MultiSeriesArimaModel]) -> None:
+def mlflow_arima_log_model(arima_model: Union[ArimaModel, MultiSeriesArimaModel],
```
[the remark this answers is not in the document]
lu-wang-dl: Just keep it since we may need to use the corresponding flavors in mlflow in the future.

runtime/databricks/automl_runtime/forecast/model.py
lu-wang-dl: To goal to add the infered signature is to make `predict` function work from mlflow side. We need to use `predict` to get the right output format.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -40,7 +42,7 @@
 }
 
 
-class AbstractArimaModel(ABC, mlflow.pyfunc.PythonModel):
+class AbstractArimaModel(ForecastModel):
```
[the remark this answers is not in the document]
lu-wang-dl: Sure.

runtime/databricks/automl_runtime/forecast/model.py
wenfeiy-db: Oh I see. I missed the `sample_input` in the line. It makes sense then that we call `predict` to get a sample prediction. 
Is there a way to test in unit tests if the `predict` function work from mlflow side?

wenfeiy-db (approved):

wenfeiy-db:
BTW could we prohibit merging if we have failed checks? It seems that I can still merge a PR when the unit test checking is failed.

yxiong:
> BTW could we prohibit merging if we have failed checks? It seems that I can still merge a PR when the unit test checking is failed.

+1. Please double check the unit test failure.

lu-wang-dl:
Disabled the merging when test is failed.

runtime/databricks/automl_runtime/forecast/model.py
lu-wang-dl: mlflow has that unit test, I don't think we need to verify it on our side. We only need to make sure our predict function works properly and assume that `infer_signature` works.

PR #32, opened 2022-02-09
[ML-19989] Add `impute_method` parameter in DatetimeTransformer

yxiong:
Main change to this PR: add an additional variable, `impute_method` to the datetime transformers to decide how to fill in the missing values. (Without this change, we will fill `1970-01-01`, which will still be the default after this change.)

Secondary change: refactored the code a little to keep common logic in the base class, and use two small abstractmethod override to modify behaviors in subclasses.

The universe counterpart PR is https://github.com/databricks/universe/pull/148207, which is e2e tested in https://e2-dogfood.staging.cloud.databricks.com/?o=6051921418418893#notebook/2133117006757276.

### DIFF
--- a/runtime/databricks/automl_runtime/sklearn/base_datetime_transformer.py
+++ b/runtime/databricks/automl_runtime/sklearn/base_datetime_transformer.py
@@ -71,8 +71,8 @@ class BaseDatetimeTransformer(ABC, TransformerMixin, BaseEstimator):
 
     def fit(self, X, y=None):
         """
-        Do nothing and return the estimator unchanged.
-        This method is just there to implement the usual API and hence work in pipelines.
+        Find the `mean`, `median` or `most_frequent` values of the input data
+        if it is needed for imputation.
         """
         x = self._to_datetime(X.iloc[:, 0])
         if self.impute_method == "mean":
--- a/runtime/databricks/automl_runtime/sklearn/base_datetime_transformer.py
+++ b/runtime/databricks/automl_runtime/sklearn/base_datetime_transformer.py
@@ -36,16 +36,57 @@ class BaseDatetimeTransformer(ABC, TransformerMixin, BaseEstimator):
     DAYS_IN_YEAR = 366  # Account for leap years
     WEEKEND_START = 5
 
+    def __init__(self, impute_method=EPOCH):
+        """Create a `BaseDatetimeTransformer`.
+
+        Parameters
+        ----------
+        impute_method: method used for imputation
+        """
+        self.impute_method = impute_method
+        if impute_method not in ("mean", "median", "most_frequent"):
+            self.impute_value = pd.to_datetime(self.impute_method)
+
+    def _to_datetime(self, X):
+        pass
+
+    def _include_timestamp(self):
+        pass
+
     def fit(self, X, y=None):
         """
         Do nothing and return the estimator unchanged.
         This method is just there to implement the usual API and hence work in pipelines.
         """
+        x = self._to_datetime(X.iloc[:, 0])
+        if self.impute_method == "mean":
+            self.impute_value = x.mean(skipna=True)
+        elif self.impute_method == "median":
+            self.impute_value = x.median(skipna=True)
+        elif self.impute_method == "most_frequent":
+            self.impute_value = x.mode(dropna=True)[0]
         return self
 
-    @abstractmethod
     def transform(self, X):
-        pass  # pragma: no cover
+        """
+        Transform timestamp data to datetime features.
+
+        Parameters
+        ----------
+        X : pd.DataFrame of shape (n_samples, 1)
+            The only column is either a timestamp column or string column with
+            datetime values encoded in ISO 8601 format.
+
+        Returns
+        -------
+        X_tr : pd.DataFrame of shape (n_samples, 13)
+            Transformed features.
+        """
+        # Convert column to datetime if data type is string and standardize to UTC
+        X.iloc[:, 0] = self._to_datetime(X.iloc[:, 0])
+        X = X.fillna(self.impute_value)
+
+        return self._generate_datetime_features(X, include_timestamp=self._include_timestamp())
 
     @staticmethod
     def _cyclic_transform(unit, period):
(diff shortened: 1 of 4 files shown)


runtime/databricks/automl_runtime/sklearn/base_datetime_transformer.py
```diff
@@ -36,16 +40,69 @@ class BaseDatetimeTransformer(ABC, TransformerMixin, BaseEstimator):
...
+
+        Parameters
+        ----------
+        impute_method: method used for imputation, with the following options:
+            'mean', 'median', 'most_frequent', or a string that can be converted to
+            a date or timestamp by `pd.to_datetime` function.
+            If not specified, the default is "1970-01-01" (unix epoch).
+        """
+        if not impute_method:
+            impute_method = self.EPOCH
```
lu-wang-dl, line 54: EPOCH is not a method. What is the purpose to set the value to EPOCH?

lu-wang-dl: Should we consider to change the default method?

lu-wang-dl (changes requested):
If the main purpose of the transformer is to impute the null values now, do we want to change the name to something like `BaseDatetimeImputer`?

runtime/databricks/automl_runtime/sklearn/base_datetime_transformer.py
```diff
@@ -36,16 +40,69 @@ class BaseDatetimeTransformer(ABC, TransformerMixin, BaseEstimator):
...
+        pass  # pragma: no cover
+
+    @abstractmethod
+    def _include_timestamp(self):
+        """Whether the output should include timestamp features."""
+        pass  # pragma: no cover
+
     def fit(self, X, y=None):
         """
         Do nothing and return the estimator unchanged.
```
wenfeiy-db, line 74: Shall we change this docstring? It now calculates the imputation value when calling fit.

wenfeiy-db:
> If the main purpose of the transformer is to impute the null values now, do we want to change the name to something like `BaseDatetimeImputer`?

I think the transformation done by `pd.to_datetime` is also a major transformation. So simply calling the class imputer might be confusing. We might want to list what transformations we do in the class docstirng.

runtime/databricks/automl_runtime/sklearn/base_datetime_transformer.py
lu-wang-dl: Then do we want to put everythin into one transformer or create two?

yxiong:
> > If the main purpose of the transformer is to impute the null values now, do we want to change the name to something like BaseDatetimeImputer?

> I think the transformation done by pd.to_datetime is also a major one. So simply calling the class imputer might be confusing. We might want to list what transformations we do in the class docstirng.

Right, I would say the main purpose is still transform, or featurization, with impute as a side effect. Logically it makes sense to split them into two, but implementation-wise it's awkward (because we first need `to_datetime` to convert data to floating number to make it easier to impute with mean, median, etc.). In that case I prefer to keep it as is, until the class acquires more functionalities to make it more appealing to split. WDYT?

runtime/databricks/automl_runtime/sklearn/base_datetime_transformer.py
yxiong: I agree here the "method" is not the best naming, as users can pass in either "mean" or a constant. `EPOCH` is a constant `1970-01-01`, which was the default value.

For default method, I kept `1970-01-01` to be backward compatible, and feel it's not a very bad option. It essentially signifies the date is missing, and our featurization will tell it apart from regular dates. With that said, I think "median" or "most_frequent" are equally good. Let me know what you feel is the best default.

runtime/databricks/automl_runtime/sklearn/base_datetime_transformer.py
yxiong: Updated the docstring. I'm still leaning towards to keep things in one transformer for now due to technical convenience (see my other comments).

yxiong:
Thanks for the review, PTAL.

wenfeiy-db (approved):
LGTM. Only having another minor comment on docstring. Could we list what transformations (i.e. to_datetime, impute & converting datetime values to datetime features) we do in the docstring of the class or `transform` method? I felt this will help us to keep track of the transformations as we add new ones in the future and also make it easier to see when would be a suitable time to split classes.

runtime/databricks/automl_runtime/sklearn/base_datetime_transformer.py
wenfeiy-db: Agree on keeping it in one for now. We need the `pd.to_datetime` first, and then do imputation, and finally calculate the datetime features. Seems need to split to at least three classes, which might be an overkill.

runtime/databricks/automl_runtime/sklearn/base_datetime_transformer.py
```diff
@@ -36,16 +40,74 @@ class BaseDatetimeTransformer(ABC, TransformerMixin, BaseEstimator):
...
+        ----------
+        impute_method: method used for imputation, with the following options:
+            'mean', 'median', 'most_frequent', or a string that can be converted to
+            a date or timestamp by `pd.to_datetime` function.
+            If not specified, the default is "1970-01-01" (unix epoch).
+        """
+        if not impute_method:
+            impute_method = self.EPOCH
+        self.impute_method = impute_method
+        if impute_method in ("mean", "median", "most_frequent"):
```
lu-wang-dl, line 56: Do we plan to support the constant imputation method?

runtime/databricks/automl_runtime/sklearn/base_datetime_transformer.py
lu-wang-dl: So if the impute_method is a string, it means is it a constant imputation? Then I would suggest to change the name to make it more clear. Or add a const imputation method with extra input.

lu-wang-dl:
Do we really want to put everything into one transformer? Could we define another transformer for imputation?

lu-wang-dl:
> 

If we mixed multiple things in one transformer, it is convenient for us but the users may not be able to understand the transformers easily and it is also not easy to explain the model.

yxiong:
Discussed with Lu offline and we decided it's better to move the imputation logic to a dedicated class. Dropping this PR. New PR at https://github.com/databricks/automl/pull/33.

PR #33, opened 2022-02-10
[ML-19989] Implement a `DatetimeImputer`

yxiong:


runtime/tests/automl_runtime/sklearn/datetime_imputer_test.py
```diff
@@ -0,0 +1,61 @@
...
+class TestDatetimeImputer(unittest.TestCase):
+    def get_test_df(self, fill_value_1=None, fill_value_2=None):
+        return pd.DataFrame({
+            "f1": ["2021-01-01", fill_value_1, "2021-01-03", "2021-01-05", "2021-01-07", "2021-01-07"],
+            "f2": [f"2022-01-01 00:0{i}" for i in [1,3,5,7,7]] + [fill_value_2]})
+
+    def to_datetime(self, X):
+        return pd.DataFrame({
+            col_name: pd.to_datetime(col_value)
+            for col_name, col_value in X.iteritems()})
```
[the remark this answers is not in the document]
yxiong: Ah, I didn't know this trick. Thanks for the suggestion. Updated.

runtime/databricks/automl_runtime/sklearn/datetime_imputer.py
```diff
@@ -0,0 +1,77 @@
...
+# limitations under the License.
+#
+
+import pandas as pd
+from sklearn.base import TransformerMixin, BaseEstimator
+
+class DatetimeImputer(TransformerMixin, BaseEstimator):
+    """Imputer for date and timestamp data."""
+
+    def __init__(self, strategy='median', fill_value=None):
```
Loquats, line 23: How about we default to mean instead of median? I think they're both reasonable defaults, but it would be easier for us and users to remember if we keep it consistent with the default method used for numerical values. Otherwise this would be the only data type that default to median.

We discussed this briefly during design
https://docs.google.com/spreadsheets/d/1Tz5sIn-FF7dY-comdAtkr4bU15hpxnpbn6EOZgvBP8A/edit#gid=1056641222

runtime/databricks/automl_runtime/sklearn/datetime_imputer.py
```diff
@@ -0,0 +1,77 @@
...
+        """Convert the input to datetime object and then impute the missing values.
+
+        Parameters
+        ----------
+        X: a pandas DataFrame whose values are date or timestamp.
+        """
+        return pd.DataFrame({
+            col_name: pd.to_datetime(col_value, errors="coerce").fillna(
+                self.fill_values[col_name])
+            for col_name, col_value in X.iteritems()
```
Loquats, line 74: Let's try to avoid looping through the columns here, to allow pandas to possibly do this more efficiently. `to_datetime` can operate on the entire dataframe, and I think we can also do that for `fillna` by specifying the `value` argument as a dict or Series:
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.fillna.html#pandas-dataframe-fillna

yxiong: Good point, updated.

runtime/databricks/automl_runtime/sklearn/datetime_imputer.py
yxiong: The reason I think median makes more sense is when the input is date (not timestamp), it is like a discrete, ordered datatype. Using median will still get us a date, but using mean will return a timestamp in between dates. WDYT?

yxiong: Chatted offline and updated the strategy to `mean`.

PR #34, opened 2022-02-18
Add unit for seasonality value

wenfeiy-db:
Prophet's seasonality value is always in days, while the `generate_cutoffs` function previously uses the input frequency to calculate `seasonality_timedelta`. This will cause an incorrect and super large initial period.

This PR fixes the above issue.

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/utils.py
+++ b/runtime/databricks/automl_runtime/forecast/utils.py
@@ -26,7 +26,7 @@ def generate_cutoffs(df: pd.DataFrame, horizon: int, unit: str,
     :param df: pd.DataFrame of the historical data.
     :param horizon: int number of time into the future for forecasting.
     :param unit: frequency of the timeseries, which must be a pandas offset alias.
-    :param seasonal_period: length of the seasonality period in days.
+    :param seasonal_period: length of the seasonality period in given frequency.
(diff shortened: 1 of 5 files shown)


runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -20,21 +20,22 @@
 
 
 def generate_cutoffs(df: pd.DataFrame, horizon: int, unit: str,
-                     seasonal_period: int, num_folds: int) -> List[pd.Timestamp]:
```
lu-wang-dl, line 23: Should we make the default value to "D" to simplify the function calls in other places?

lu-wang-dl (approved):
LGTM. Just one comment.

runtime/databricks/automl_runtime/forecast/utils.py
wenfeiy-db: I'm not sure if it makes sense to have "D" as the default value, because it's just Prophet always reporting the seasonality values in days. It might be more logical to make the `seasonal_unit` optional, and if it's not set, we use the `unit` from the required field. WDYT?

lu-wang-dl (approved):
LGTM. Just one comment.

PR #39, opened 2022-03-28
Add option include_history in predict_timeseries for forecasting model

lu-wang-dl:
Add option include_history so that users can get the predicted data simply if they want.

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/prophet/model.py
+++ b/runtime/databricks/automl_runtime/forecast/prophet/model.py
@@ -83,31 +83,39 @@ class ProphetModel(ForecastModel):
         from prophet.serialize import model_from_json
         return model_from_json(self._model_json)
 
-    def _make_future_dataframe(self, horizon: int) -> pd.DataFrame:
+    def _make_future_dataframe(self, horizon: int, include_history: bool = True) -> pd.DataFrame:
         """
         Generate future dataframe by calling the API from prophet
         :param horizon: Int number of periods to forecast forward.
+        :param include_history: Boolean to include the historical dates in the data
+            frame for predictions.
         :return: pd.Dataframe that extends forward from the end of self.history for the
         requested number of periods.
         """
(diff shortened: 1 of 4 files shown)


runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -162,42 +170,53 @@ def model(self, id: str) -> prophet.forecaster.Prophet:
...
+            frame for predictions.
         :return: pd.Dataframe that extends forward from the end of self.history for the
         requested number of periods.
         """
-        start_time = self._timeseries_starts[id]
         end_time = pd.Timestamp(self._timeseries_end)
+        if include_history:
+            start_time = self._timeseries_starts[id]
+        else:
+            start_time = end_time + pd.Timedelta(value=1, unit=self._frequency)
```
wenfeiy-db, line 187: Why do we use a same `end_time` for all time series in prophet? It doesn't sound quite accurate.

wenfeiy-db (approved):
The part for adding `include_history` LGTM. But I'm having question about why we only use a single `end_time` for forecasting in prophet. This don't need to be addressed in the PR though.

lu-wang-dl:
That is based on one assumption, we want to predict the future data.  For multi-series, all series ends at the same time. If not, it means it has missing data. For example, for two series in one data, one is ended at `02-01-2022` and the other is ended at `02-02-2022` and horizon is 1. We want to predict the data at `02-03-2022` since it is the future, but not `02-02-2022` for the first time series.

wenfeiy-db:
> That is based on one assumption, we want to predict the future data. For multi-series, all series ends at the same time. If not, it means it has missing data. For example, for two series in one data, one is ended at `02-01-2022` and the other is ended at `02-02-2022` and horizon is 1. We want to predict the data at `02-03-2022` since it is the future, but not `02-02-2022` for the first time series.

That makes sense.  Then we probably want to modify the ARIMA wrapper to use the same logic. Currently it's using different end time for each time series.

PR #44, opened 2022-04-29
Make the parameters of early stop function configurable

yxiong:


Loquats (approved):
LGTM. Looks like the implementation of `early_stop_fn` is unchanged , just got wrapped in `get_early_stop_fn`.

Since this is a breaking change (different function exposed), let's be careful about updating the commit version used in automl DUST https://src.dev.databricks.com/databricks/universe@master/-/blob/spark/images/extensions/ml/tests/dust/AutoMLUtils.scala?L18

PR #49, opened 2022-05-05
Add test for string type time columns

lu-wang-dl:
- Fix a bug in pmdarima/training.py
- Add test for string type time column in training and inference for pmdarima and prophet.

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/pmdarima/training.py
+++ b/runtime/databricks/automl_runtime/forecast/pmdarima/training.py
@@ -60,7 +60,7 @@ class ArimaEstimator:
         history_pd["ds"] = pd.to_datetime(history_pd["ds"])
 
         # Check if the time has consistent frequency
-        self._validate_ds_freq(df, self._frequency_unit)
+        self._validate_ds_freq(history_pd, self._frequency_unit)
 
         # Impute missing time steps
         history_pd = self._fill_missing_time_steps(history_pd, self._frequency_unit)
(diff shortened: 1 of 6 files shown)


wenfeiy-db (approved):
Changes LGTM!
I'm wondering the expected behavior for string time columns. With changes in this PR, any string can be converted by `pd.to_datetime` will work for arima. What about the prophet? I have the impression that they also use `pd.to_datetime` internally, but I'm not sure  if it'll be changed or if it's consistently used in prophet.
Generally I think we need to make a consistent criteria for the string time col for arima and prophet, and we'd better make the criteria clear and force it ourselves. This shouldn't block this PR of course.

lu-wang-dl:
By default, prophet is using `pd. to_datetime ` in prophet call: https://github.com/facebook/prophet/blob/b7d5ec448177b7ac2abf25cc3259cb55e4c44069/python/prophet/forecaster.py#L267

wenfeiy-db:
> By default, prophet is using `pd. to_datetime ` in prophet call: https://github.com/facebook/prophet/blob/b7d5ec448177b7ac2abf25cc3259cb55e4c44069/python/prophet/forecaster.py#L267

Thanks for the pointer! I have the impression as well, but what I mean is we should guarantee this ourselves instead of relying on the implementation details of other projects. This could make the expected behavior against string column clearer as well as avoid breaking changes introduced by external project. wdyt?

lu-wang-dl:
Sure. Let me also add the type conversion in prophet function too to make sure we will have the same one for all models.

PR #58, opened 2022-06-03
Fix multi id prediction error

lu-wang-dl:


wenfeiy-db (approved):

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -33,7 +33,7 @@
             "pip": [
                 f"prophet=={prophet.__version__}",
                 f"cloudpickle=={cloudpickle.__version__}",
-                f"databricks-automl-runtime==0.2.5",
```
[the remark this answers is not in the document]
yxiong: +1

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -262,7 +264,12 @@ def predict(self, context: mlflow.pyfunc.model.PythonModelContext, model_input:
         test_df = model_input.copy()
         test_df["ts_id"] = test_df[self._id_cols].astype(str).agg('-'.join, axis=1)
         test_df.rename(columns={self._time_col: "ds"}, inplace=True)
-        predict_df = test_df.groupby("ts_id").apply(lambda df: self.model(df.name[0]).predict(df)).reset_index()
+
+        def model_prediction(df):
+            model = self.model(df.name)
+            if model:
+                return model.predict(df)
+        predict_df = test_df.groupby("ts_id").apply(lambda df: model_prediction(df)).reset_index()
```
yxiong, line 272: What would the `apply` do if the `model_prediction` returns `None` (not in the `if` branch)?

runtime/databricks/automl_runtime/forecast/prophet/model.py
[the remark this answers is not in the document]
lu-wang-dl: The issue to use the version from version.py is that in the env file of the model will include a package that not on pypi. It may cause some issue for our testing with the main branch of the repo. But it is fine on the released ones.

runtime/databricks/automl_runtime/forecast/prophet/model.py
lu-wang-dl: It is fine in that case, let me add a unit test for this case.

yxiong (approved):

runtime/databricks/automl_runtime/forecast/prophet/model.py
lu-wang-dl: Confirmed in the unit test, it will return nan.

PR #59, opened 2022-06-16
Fix MultiSeriesProphetModel ts_id key error with one-row dataframe

wenfeiy-db:
If we use a loaded `MultiSeriesProphetModel` from mlflow and use a one-row dataframe as input of `predict`, then it will throw a KeyError for `ts_id`. For some reason the `ts_id` is not kept as an index after `test_df.groupby("ts_id").apply(lambda df: model_prediction(df))` when `test_df` only has one row. To fix this, we explicitly assign ts_id in `model_prediction`.

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/prophet/model.py
+++ b/runtime/databricks/automl_runtime/forecast/prophet/model.py
@@ -270,9 +270,13 @@ class MultiSeriesProphetModel(ProphetModel):
             model = self.model(df.name)
             if model:
                 predicts = model.predict(df)
+                # We have to explicitly assign the ts_id to avoid KeyError when model_input
+                # only has one row. For multi-rows model_input, the ts_id will be kept as index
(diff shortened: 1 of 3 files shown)


runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -269,8 +269,10 @@ def predict(self, context: mlflow.pyfunc.model.PythonModelContext, model_input:
         def model_prediction(df):
             model = self.model(df.name)
             if model:
-                return model.predict(df)
-        predict_df = test_df.groupby("ts_id").apply(lambda df: model_prediction(df)).reset_index()
+                predicts = model.predict(df)
+                predicts["ts_id"] = df.name
```
panchalhp-db, line 273: Can we add a comment on why this is required?

panchalhp-db (approved):

PR #66, opened 2022-06-27
Update python env and packages

panchalhp-db:
Update environment to be on par with the latest ML Runtime version

### DIFF
--- a/.github/workflows/runtime-test.yml
+++ b/.github/workflows/runtime-test.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8]
+        python-version: [3.9]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


runtime/test-environment.txt
```diff
@@ -3,4 +3,4 @@
 # * Keep dependencies sorted.
 pytest~=6.2.1
 pytest-cov~=2.11.1
-testfixtures~=6.17.1
```
wenfeiy-db, line 6: Is this no longer needed?

.github/workflows/runtime-test.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8]
```
wenfeiy-db, line 15: Do we want to maintain the package for different python versions?

wenfeiy-db (commented):
Seems that we also need to update the check required for merging the PR. Currently it still expect a 3.8 build?

.github/workflows/runtime-test.yml
panchalhp-db: Shouldn't the other branches have their own build file with 3.8?

panchalhp-db:
> Seems that we also need to update the check required for merging the PR. Currently it still expect a 3.8 build?

@wenfeiy-db yeah seems like it. I don't have permission to change the repo settings, do you have permissions?

.github/workflows/runtime-test.yml
wenfeiy-db: Hmmm good question. I think it should. not 100% sure though

panchalhp-db:
> > > Seems that we also need to update the check required for merging the PR. Currently it still expect a 3.8 build?
> 
> > 
> 
> > @wenfeiy-db yeah seems like it. I don't have permission to change the repo settings, do you have permissions?
> 
> 
> 
> Hmm I don't think so. On the settings page, I don't see the "Code and automation" section, on which we can set the protected branch rules ([github doc](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule)).
> 
> We might need Lu's help this....

Okay yeah for now maybe I'll keep both the python envs in the matrix.

runtime/test-environment.txt
panchalhp-db: Nope, we aren't using this in this repo

.github/workflows/runtime-test.yml
panchalhp-db: yeah I think older release branches will build with the version of build file on their branch so we should be okay

yxiong: Looks like the conclusion is we should only keep `3.9` in master, right? The current update has both `3.8` and `3.9`.

yxiong (approved):

.github/workflows/runtime-test.yml
panchalhp-db: @yxiong yep because we'll probably need Lu to change the repo settings to allow the PR to merge without the 3.8 job. Currently without it, the PR shows check pending with expected results for 3.8

PR #85, opened 2022-09-14
[ES-412626] Call pd.to_datetime on start_ds and end_ds

maddiedawson:


wenfeiy-db (approved):
LGTM!
Could we add a unit test to make sure the forecast doesn't fail when the input start_ds is a datetime.date type?

PR #86, opened 2022-09-23
[ML-25376] Grant non-string column names to customized transformers.

shitaoli-db:
[scikit-learn/sklearn/compose/_column_transformer.py at 36958fb240fbe435673a9e3c52e769f01f36bec0 · scikit-learn/scikit-learn](https://github.com/scikit-learn/scikit-learn/blame/36958fb240fbe435673a9e3c52e769f01f36bec0/sklearn/compose/_column_transformer.py#L329-L340)

Sk learn 1.0 has some wierd logic to use str name on column to reverse look up on the column transformer, this would cause trouble if all our column names are using same string.


runtime/databricks/automl_runtime/sklearn/base_datetime_transformer.py
```diff
@@ -117,5 +117,8 @@ def _generate_datetime_features(cls, X, include_timestamp=True):
 
         for holiday_calendar in [holidays.UnitedStates(), holidays.EuropeanCentralBank()]:
             features.append(X[col].apply(lambda datetime: datetime in holiday_calendar))
-
-        return pd.concat(features, axis=1)
+        ans = pd.concat(features, axis=1)
+        # Give non-string column names to avoid duplicated feature names, sklearn 1.0 does not work on duplicated
+        # feature nanmes.
+        ans.columns = range(len(ans.columns))
```
yxiong, line 123: Is this the only breakage to scikit-learn 1.0? Can we update https://github.com/databricks/automl/blob/64af351986dd6fb0798ef09129b8af0e6fbb92ae/runtime/environment.txt#L12 to ensure all the tests passed?

shitaoli-db: Done. Also added a new test case.

yxiong (approved):

PR #88, opened 2022-09-26
[ML-22031] Support changes made in universe related to datetime parsing

rithwik-db:
This is a change necessary to reconcile the difference in how datetime parsing happens. Look at https://github.com/databricks/universe/pull/211250 for additional information.

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/pmdarima/model.py
+++ b/runtime/databricks/automl_runtime/forecast/pmdarima/model.py
@@ -145,7 +145,7 @@ class ArimaModel(AbstractArimaModel):
 
     def _predict_impl(self, input_df: pd.DataFrame) -> pd.DataFrame:
         df = input_df.rename(columns={self._time_col: "ds"})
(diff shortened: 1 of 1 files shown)


runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -145,7 +145,7 @@ def predict(self, context: mlflow.pyfunc.model.PythonModelContext, model_input:
 
     def _predict_impl(self, input_df: pd.DataFrame) -> pd.DataFrame:
         df = input_df.rename(columns={self._time_col: "ds"})
-        df["ds"] = pd.to_datetime(df["ds"])
+        df["ds"] = pd.to_datetime(df["ds"], infer_datetime_format=True, errors="coerce")
```
maddiedawson, line 148: Actually, on second thought, I think that we don't want `errors="coerce"` during inference so that the error is surfaced to the user. Let's remove that but keep the `infer_datetime_format=True`

maddiedawson (approved):

PR #99, opened 2022-11-09
[ML-27512] Pass exogenous var to auto_arima

maddiedawson:
Modify Arima APIs to use regressors/exogenous variables during training and inference. If a model is trained with regressors, the regressors must be passed in during inference.

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/pmdarima/model.py
+++ b/runtime/databricks/automl_runtime/forecast/pmdarima/model.py
@@ -117,7 +117,12 @@ class ArimaModel(AbstractArimaModel):
         """
         return pickle.loads(self._pickled_model)
 
-    def predict_timeseries(self, horizon: int = None, include_history: bool = True) -> pd.DataFrame:
+    # TODO(maddie.dawson): Remove X from this signature?
+    def predict_timeseries(
+        self,
+        horizon: int = None,
+        X: pd.DataFrame = None,
+        include_history: bool = True) -> pd.DataFrame:
         """
         Predict target column for given horizon_timedelta and history data.
         :param horizon: int number of periods to forecast forward.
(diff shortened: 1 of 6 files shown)


runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -177,18 +183,30 @@ def _predict_impl(self, input_df: pd.DataFrame) -> pd.DataFrame:
         # Out-of-sample prediction if needed
         horizon = calculate_period_differences(self._end_ds, max(df["ds"]), self._frequency)
         if horizon > 0:
-            future_pd = self._forecast(horizon)
+            # Forward fill missing time steps
+            X_future = df[df["ds"] > self._end_ds].set_index("ds").resample(rule=OFFSET_ALIAS_MAP[self._frequency]).pad()
```
lu-wang-dl, line 187: Maybe later, we could handle this in the data preprocessing step.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -283,7 +310,11 @@ def make_future_dataframe(
         )
         return future_df
 
-    def predict_timeseries(self, horizon: int = None, include_history: bool = True) -> pd.DataFrame:
+    def predict_timeseries(
```
lu-wang-dl, line 313: Should we create a new function instead of changing this one?

lu-wang-dl (commented):
Overall LGTM. Left few comments and wait for fixing the tests.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
maddiedawson: I made the df optional

lu-wang-dl (approved):

PR #104, opened 2022-12-21
Reverted back the fill missing timestamp to make the behavior aligned with non-monthly data.

shitaoli-db:


wenfeiy-db (approved):

wenfeiy-db:
LGTM - why don't we do a simple revert though?

shitaoli-db:
> LGTM - why don't we do a simple revert though?

To make monthly work on non-filling situation.

PR #107, opened 2023-01-04
Add make_future_dataframe function for forecast model

lu-wang-dl:
- Update the id as tuple instead of string
- Add make_future_dataframe function
- Update test env package version

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/pmdarima/model.py
+++ b/runtime/databricks/automl_runtime/forecast/pmdarima/model.py
@@ -227,7 +227,7 @@ class MultiSeriesArimaModel(AbstractArimaModel):
         :param horizon: int number of periods to forecast forward.
         :param frequency: the frequency of the time series
         :param start_ds_dict: the dictionary of the starting time of each time series in training data.
-        :param end_ds_dict: the dictionary of the starting time of each time series  in training data.
+        :param end_ds_dict: the dictionary of the end time of each time series in training data.
         :param time_col: the column name of the time column
         :param id_cols: the column names of the identity columns for multi-series time series
         """
@@ -279,7 +279,7 @@ class MultiSeriesArimaModel(AbstractArimaModel):
             frequency=self._frequency,
             include_history=include_history,
             groups=groups,
-            group_names=self._id_cols
+            identity_column_names=self._id_cols
         )
         return future_df
 
--- a/runtime/databricks/automl_runtime/forecast/pmdarima/model.py
+++ b/runtime/databricks/automl_runtime/forecast/pmdarima/model.py
@@ -15,7 +15,7 @@
 #
 import pickle
 from abc import abstractmethod
-from typing import List, Dict, Union
+from typing import List, Dict, Tuple, Optional, Union
 
 import pandas as pd
 import mlflow
@@ -25,7 +25,8 @@ from mlflow.protos.databricks_pb2 import INVALID_PARAMETER_VALUE
 
 from databricks.automl_runtime.forecast import OFFSET_ALIAS_MAP, DATE_OFFSET_KEYWORD_MAP
 from databricks.automl_runtime.forecast.model import ForecastModel, mlflow_forecast_log_model
-from databricks.automl_runtime.forecast.utils import calculate_period_differences, is_frequency_consistency
+from databricks.automl_runtime.forecast.utils import calculate_period_differences, is_frequency_consistency, \
+    make_future_dataframe
 
 
 ARIMA_CONDA_ENV = {
@@ -217,8 +218,8 @@ class MultiSeriesArimaModel(AbstractArimaModel):
     ARIMA mlflow model wrapper for multivariate forecasting.
     """
 
-    def __init__(self, pickled_model_dict: Dict[str, bytes], horizon: int, frequency: str,
-                 start_ds_dict: Dict[str, pd.Timestamp], end_ds_dict: Dict[str, pd.Timestamp],
+    def __init__(self, pickled_model_dict: Dict[Tuple, bytes], horizon: int, frequency: str,
+                 start_ds_dict: Dict[Tuple, pd.Timestamp], end_ds_dict: Dict[Tuple, pd.Timestamp],
                  time_col: str, id_cols: List[str]) -> None:
         """
         Initialize the mlflow Python model wrapper for multiseries ARIMA.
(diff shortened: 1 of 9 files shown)


runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -14,14 +14,77 @@
...
+) -> pd.DataFrame:
+    if groups is None:
+        return make_single_future_dataframe(start_time, end_time, horizon, frequency)
+
+    future_df_list = []
+    for group in groups:
+        if type(start_time) is dict:
+            group_start_time = start_time[group]
+        else:
+            group_start_time = start_time
```
wenfeiy-db, line 42: In which cases we'll have a single timestamp as `start_time` when we have multiple series? I thought we always use a dict for multi series cases.

runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -14,14 +14,77 @@
...
 _logger = logging.getLogger(__name__)
 
+def make_future_dataframe(
+        start_time: Union[pd.Timestamp, Dict[Tuple, pd.Timestamp]],
+        end_time: Union[pd.Timestamp, Dict[Tuple, pd.Timestamp]],
+        horizon: int,
+        frequency: str,
+        include_history: bool = True,
+        groups: List[Tuple] = None,
+        group_names: List[str] = None,
```
wenfeiy-db, line 32: Would it be more readable if we use a `Dict[str, Tuple]` here, which maps from group name to group (i.e. tuple)? This way we don't have to bother keeping the two lists to be same order.

Also, could we add docstr for this function? It took me a while to understand what's groups and what's group_names.

runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -14,14 +14,77 @@
...
+from typing import Dict, List, Optional, Tuple, Union
 from databricks.automl_runtime.forecast import DATE_OFFSET_KEYWORD_MAP,\
     QUATERLY_OFFSET_ALIAS, NON_DAILY_OFFSET_ALIAS, OFFSET_ALIAS_MAP, PERIOD_ALIAS_MAP
 
 import pandas as pd
 
 _logger = logging.getLogger(__name__)
 
+def make_future_dataframe(
+        start_time: Union[pd.Timestamp, Dict[Tuple, pd.Timestamp]],
```
wenfeiy-db, line 26: Shall we add unit tests for the two util functions?

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -176,30 +177,53 @@ def model(self, id: str) -> Optional[prophet.forecaster.Prophet]:
             return model_from_json(self._model_json[id])
         return None
 
-    def _make_future_dataframe(self, id: str, horizon: int, include_history: bool = True) -> pd.DataFrame:
+    def make_future_dataframe(
```
wenfeiy-db, line 180: I'm a bit confused why we need both this method and the util.make_future_dataframe. Is this method used anywhere?

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -176,30 +177,53 @@ def model(self, id: str) -> Optional[prophet.forecaster.Prophet]:
...
+            end_time=end_time,
+            horizon=horizon,
+            frequency=self._frequency,
+            include_history=include_history,
+            groups=groups,
+            group_names=self._id_cols
+        )
+        return future_df
+
+    def _make_future_dataframe(self, id: Tuple, horizon: int, include_history: bool = True) -> pd.DataFrame:
```
wenfeiy-db, line 215: Shall we change the method `_make_future_dataframe` in the univariate model wrapper `ProphetModel` as well to use our own util function for consistency? It's still calling self.model().make_future_dataframe.

runtime/databricks/automl_runtime/forecast/prophet/model.py
lu-wang-dl: Adding the function to the model because the model has the infos like start and end time, horizon, frequencies. The actual implementation is in util so that the two models can use the same one.

runtime/databricks/automl_runtime/forecast/utils.py
lu-wang-dl: We don't have this use case. Add this part only to avoid unknown error when user try to call this function.

runtime/databricks/automl_runtime/forecast/prophet/model.py
lu-wang-dl: Good point. I will make a follow up on this. Currently `ProphetModel` does not contain time series starts and ends. So it is not simple to use this util function.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -281,7 +318,7 @@ def predict(self, context: mlflow.pyfunc.model.PythonModelContext, model_input:
         """
         self._validate_cols(model_input, self._id_cols + [self._time_col])
         df = model_input.copy()
-        df["ts_id"] = df[self._id_cols].apply(lambda r: "-".join(r.values.astype(str)), axis=1)
+        df["ts_id"] = df[self._id_cols].apply(tuple, axis=1)
```
shitaoli-db, line 321: Nit: Is it possible to use a UUID based or at least a constant for this "ts-id", since this may have trouble if user data frame has column called "ts_id". And use hard code string can sometimes write bugs in future development.

shitaoli-db (commented):
Generally LGTM.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
lu-wang-dl: This is also a issue from prophet which reserved multiple column names. We can change this as a followup.

runtime/databricks/automl_runtime/forecast/prophet/model.py
lu-wang-dl: It is designed to be used in the trial notebook for model inference with extra features. We will generate the future dataframe, join feature store tables and predict the future data.

wenfeiy-db: ahhh got it. That makes sense! Thanks for providing the context!

runtime/tests/automl_runtime/forecast/utils_test.py
```diff
@@ -219,4 +221,57 @@ def test_frequency_consistency(self):
...
+        )
+        self.assertEqual(len(future_df), 1)
+        expected_columns = { "test_date" }
+        self.assertTrue(expected_columns.issubset(set(future_df.columns)))
+
+        future_df = make_single_future_dataframe(
+            start_time=pd.to_datetime('2022-01-01'),
+            end_time=pd.to_datetime('2022-01-04'),
+            horizon=1,
+            frequency="d",
```
wenfeiy-db, line 245: Shall we cover difference frequencies?

wenfeiy-db (approved):

PR #108, opened 2023-01-05
Update test with Python 3.10

lu-wang-dl:


.github/workflows/publish-to-pypi.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8]
+        python-version: ["3.9"]
```
panchalhp-db, line 15: Should this be 3.10?

.github/workflows/runtime-test.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: ["3.9", "3.10"]
```
panchalhp-db, line 15: Do we not have any supported versions that are still using 3.8?

lu-wang-dl: We have 10.4LTS using 3.8. But even we need to update the code for 10.4, we will merge into branch-0.2.6. It should not affect the main branch, right?

panchalhp-db: Ah right, that's correct.

panchalhp-db (approved):

PR #109, opened 2023-01-05
Add PandasSimpleImputer

lu-wang-dl:
A wrapper of SimpleImputer with output as pandas dataframe.


runtime/databricks/automl_runtime/sklearn/pandas_simple_imputer.py
```diff
@@ -0,0 +1,55 @@
...
+#
+import pandas as pd
+from sklearn.impute import SimpleImputer
+
+class PandasSimpleImputer(SimpleImputer):
+    """
+    A wrapper of `SimpleImputer` with the support for pandas dataframe output.
+    """
+
+    def fit(self, X, y=None):
```
maddiedawson, line 24: Should we add type annotations here? Also on transform

runtime/databricks/automl_runtime/sklearn/pandas_simple_imputer.py
```diff
@@ -0,0 +1,55 @@
...
+        ----------
+
+        X : pd.DataFrame of shape = [n_samples, n_features]
+            Training dataframe, where n_samples is the number of samples
+            and n_features is the number of features.
+
+        Returns
+        -------
+
+        X_tr : pd.DataFrame of shape (n_samples, n_features_encoded)
```
maddiedawson, line 52: I think we can omit "X_tr"?

maddiedawson (approved):

runtime/databricks/automl_runtime/sklearn/pandas_simple_imputer.py
lu-wang-dl: Sure.

PR #124, opened 2023-02-24
[ML-25713] Make sure that the forecasting horizon won't exceed pandas.Timestamp.max

maddiedawson:


runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -107,14 +107,18 @@ def get_validation_horizon(df: pd.DataFrame, horizon: int, unit: str) -> int:
...
-        return horizon
-    else:
+    MIN_HORIZONS = 4  # minimum number of horizons in the dataframe
+    horizon_dateoffset = pd.DateOffset(**DATE_OFFSET_KEYWORD_MAP[unit]) * horizon
+
+    try:
+        validation_horizon = MIN_HORIZONS * horizon_dateoffset + df["ds"].min()
+    except OverflowError:
+        validation_horizon = None
+    finally:
```
panchalhp-db, line 117: does this need to be in finally? I think it can be outside the try except block.

panchalhp-db (approved):

runtime/databricks/automl_runtime/forecast/utils.py
maddiedawson: Done

panchalhp-db (approved):

PR #139, opened 2023-10-19
[ML-35749] Bump automl-runtime version for 14.2

rithwik-db:
Just bumping version

### DIFF
--- a/runtime/databricks/automl_runtime/version.py
+++ b/runtime/databricks/automl_runtime/version.py
@@ -14,4 +14,4 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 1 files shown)


jimmyxu-db (approved):

wenfeiy-db:
There seems to be no changes between 0.19 and 0.20. Maybe don't have to upgrade this package in 14.2

rithwik-db:
> There seems to be no changes between 0.19 and 0.20. Maybe don't have to upgrade this package in 14.2

@wenfeiy-db, I agree, but we might as well continue the 1:1 mapping since it doesn't really hurt, right?

PR #142, opened 2024-01-11
Fix for pandas 2.0

wenfeiy-db:
* Update environement file 
* Add py 3.11 build
* A few fixes to be compatiable with pandas 2.0

## Testing
- [x] unit tests
- [x] run automl dust in mlr

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/prophet/model.py
+++ b/runtime/databricks/automl_runtime/forecast/prophet/model.py
@@ -239,7 +239,10 @@ class MultiSeriesProphetModel(ProphetModel):
             identity_column_names=self._id_cols
         )
         future_df["ts_id"] = future_df[self._id_cols].apply(tuple, axis=1)
-        return future_df.groupby(self._id_cols).apply(lambda df: self._predict_impl(df, horizon, include_history)).reset_index()
+        print(f"predict_timeseries::future_df:\n {future_df}")
+        return future_df.groupby(self._id_cols).apply(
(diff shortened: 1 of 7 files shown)


runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -239,7 +239,10 @@ def predict_timeseries(self, horizon: int = None, include_history: bool = True)
             identity_column_names=self._id_cols
         )
         future_df["ts_id"] = future_df[self._id_cols].apply(tuple, axis=1)
-        return future_df.groupby(self._id_cols).apply(lambda df: self._predict_impl(df, horizon, include_history)).reset_index()
+        print(f"predict_timeseries::future_df:\n {future_df}")
```
wenfeiy-db, line 242: TODO: remove this

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -239,7 +239,10 @@ def predict_timeseries(self, horizon: int = None, include_history: bool = True)
...
+        return future_df.groupby(self._id_cols).apply(
+                lambda df: self._predict_impl(df, horizon, include_history)
+            ).reset_index(drop=True)
```
wenfeiy-db, line 245: Here I'm adding `drop=True` to resolve some ValueError.
I think adding `drop=True` is ok, as we have already manually added back the id cols in `_predict_impl`. But I do want to double check this with you. @lu-wang-dl

PR #145, opened 2024-07-18
[ML-42739] Add custom forecasting data splits for automl_runtime

Lanz-db:
This PR enables custom forecasting data splits. There are three main changes.
 - ArimaEstimator
   - add split_cutoff parameter
 - ProphetEstimator
   - add split_cutoff parameter
 - utils.py
   - add generate_custom_cutoff function to generate cutoffs based on given split_cutoff, for cross_validation
- [x] unit tests
- [x] manual e2e tests
- [x] add comments

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/pmdarima/training.py
+++ b/runtime/databricks/automl_runtime/forecast/pmdarima/training.py
@@ -35,7 +35,8 @@ class ArimaEstimator:
     """
 
     def __init__(self, horizon: int, frequency_unit: str, metric: str, seasonal_periods: List[int],
-                 num_folds: int = 20, max_steps: int = 150, exogenous_cols: Optional[List[str]] = None) -> None:
+                 num_folds: int = 20, max_steps: int = 150, exogenous_cols: Optional[List[str]] = None,
+                 split_cutoff: Optional[pd.Timestamp] = None) -> None:
         """
         :param horizon: Number of periods to forecast forward
         :param frequency_unit: Frequency of the time series
@@ -53,6 +54,7 @@ class ArimaEstimator:
         self._num_folds = num_folds
         self._max_steps = max_steps
         self._exogenous_cols = exogenous_cols
+        self._split_cutoff = split_cutoff
 
     def fit(self, df: pd.DataFrame) -> pd.DataFrame:
         """
@@ -88,12 +90,20 @@ class ArimaEstimator:
                 # so the minimum valid seasonality period is always 1
 
                 validation_horizon = utils.get_validation_horizon(history_pd, self._horizon, self._frequency_unit)
-                cutoffs = utils.generate_cutoffs(
-                    history_pd,
-                    horizon=validation_horizon,
-                    unit=self._frequency_unit,
-                    num_folds=self._num_folds,
-                )
+                if self._split_cutoff:
+                    cutoffs = utils.generate_custom_cutoffs(
+                        history_pd,
+                        horizon=validation_horizon,
+                        unit=self._frequency_unit,
+                        split_cutoff=self._split_cutoff
+                    )
+                else:
+                    cutoffs = utils.generate_cutoffs(
+                        history_pd,
+                        horizon=validation_horizon,
+                        unit=self._frequency_unit,
+                        num_folds=self._num_folds,
+                    )
 
                 result = self._fit_predict(history_pd, cutoffs=cutoffs, seasonal_period=m, max_steps=self._max_steps)
                 metric = result["metrics"]["smape"]
--- a/runtime/databricks/automl_runtime/forecast/pmdarima/training.py
+++ b/runtime/databricks/automl_runtime/forecast/pmdarima/training.py
@@ -35,8 +35,8 @@ class ArimaEstimator:
     """
 
     def __init__(self, horizon: int, frequency_unit: str, metric: str, seasonal_periods: List[int],
-                 num_folds: int = 20, max_steps: int = 150, exogenous_cols: Optional[List[str]] = None,
-                 split_cutoff: Optional[pd.Timestamp] = None) -> None:
+                 num_folds: int = 20, max_steps: int = 150, exogenous_cols: List[str] | None = None,
+                 split_cutoff: pd.Timestamp | None = None) -> None:
         """
         :param horizon: Number of periods to forecast forward
         :param frequency_unit: Frequency of the time series
@@ -46,6 +46,10 @@ class ArimaEstimator:
         :param max_steps: Max steps for stepwise auto_arima
         :param exogenous_cols: Optional list of column names of exogenous variables. If provided, these columns are
         used as additional features in arima model.
+        :param split_cutoff: Optional cutoff specified by user. If provided, 
+        it is the starting point of cutoffs for cross validation.
+        For tuning job, it is the cutoff between train and validate split.
+        For training job, it is the cutoff bewteen validate and test split.
         """
         self._horizon = horizon
         self._frequency_unit = OFFSET_ALIAS_MAP[frequency_unit]
--- a/runtime/databricks/automl_runtime/forecast/pmdarima/training.py
+++ b/runtime/databricks/automl_runtime/forecast/pmdarima/training.py
@@ -15,7 +15,7 @@
 #
 import logging
 import traceback
-from typing import List
+from typing import List, Optional
 
 import pandas as pd
 import pickle
@@ -35,8 +35,8 @@ class ArimaEstimator:
     """
 
     def __init__(self, horizon: int, frequency_unit: str, metric: str, seasonal_periods: List[int],
-                 num_folds: int = 20, max_steps: int = 150, exogenous_cols: List[str] | None = None,
-                 split_cutoff: pd.Timestamp | None = None) -> None:
+                 num_folds: int = 20, max_steps: int = 150, exogenous_cols: Optional[List[str]] = None,
+                 split_cutoff: Optional[pd.Timestamp] = None) -> None:
         """
         :param horizon: Number of periods to forecast forward
         :param frequency_unit: Frequency of the time series
--- a/runtime/databricks/automl_runtime/forecast/pmdarima/training.py
+++ b/runtime/databricks/automl_runtime/forecast/pmdarima/training.py
@@ -15,7 +15,7 @@
 #
 import logging
 import traceback
-from typing import List, Optional
+from typing import List
 
 import pandas as pd
 import pickle
--- a/runtime/databricks/automl_runtime/forecast/prophet/forecast.py
+++ b/runtime/databricks/automl_runtime/forecast/prophet/forecast.py
@@ -92,7 +92,7 @@ class ProphetHyperoptEstimator(ABC):
                  max_eval: int = 10, trial_timeout: int = None,
                  random_state: int = 0, is_parallel: bool = True,
                  regressors = None, 
-                 split_cutoff: pd.Timestamp | None = None, **prophet_kwargs) -> None:
+                 split_cutoff: Optional[pd.Timestamp] = None, **prophet_kwargs) -> None:
         """
         Initialization
 
--- a/runtime/databricks/automl_runtime/forecast/prophet/forecast.py
+++ b/runtime/databricks/automl_runtime/forecast/prophet/forecast.py
@@ -70,6 +70,7 @@ def _prophet_fit_predict(params: Dict[str, Any], history_pd: pd.DataFrame,
     offset_kwarg = DATE_OFFSET_KEYWORD_MAP[OFFSET_ALIAS_MAP[frequency]]
     horizon_offset = pd.DateOffset(**offset_kwarg)*horizon
     # Evaluate Metrics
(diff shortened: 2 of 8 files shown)


runtime/databricks/automl_runtime/forecast/pmdarima/training.py
```diff
@@ -35,7 +35,8 @@ class ArimaEstimator:
     """
 
     def __init__(self, horizon: int, frequency_unit: str, metric: str, seasonal_periods: List[int],
-                 num_folds: int = 20, max_steps: int = 150, exogenous_cols: Optional[List[str]] = None) -> None:
+                 num_folds: int = 20, max_steps: int = 150, exogenous_cols: Optional[List[str]] = None,
+                 split_cutoff: Optional[pd.Timestamp] = None) -> None:
```
wenfeiy-db, line 39: Can you add description of the `split_cutoff` in docstring?

runtime/databricks/automl_runtime/forecast/prophet/forecast.py
```diff
@@ -91,7 +91,8 @@ def __init__(self, horizon: int, frequency_unit: str, metric: str, interval_widt
                  algo=hyperopt.tpe.suggest, num_folds: int = 5,
                  max_eval: int = 10, trial_timeout: int = None,
                  random_state: int = 0, is_parallel: bool = True,
-                 regressors = None, **prophet_kwargs) -> None:
+                 regressors = None, 
+                 split_cutoff: Optional[pd.Timestamp] = None, **prophet_kwargs) -> None:
```
wenfeiy-db, line 95: same above

runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -187,6 +187,33 @@ def generate_cutoffs(df: pd.DataFrame, horizon: int, unit: str,
         )
     return list(reversed(result))
 
+def generate_custom_cutoffs(df: pd.DataFrame, horizon: int, unit: str,
+                     split_cutoff: pd.Timestamp) -> List[pd.Timestamp]:
+    """
+    Generate custom cutoff times for cross validation based on user-specified split cutoff.
+    Period (step size) is 1.
```
wenfeiy-db, line 194: Just a note that `period` likely needs to be exposed as an input param in the future. Ok not to do this now.

runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -187,6 +187,33 @@ def generate_cutoffs(df: pd.DataFrame, horizon: int, unit: str,
...
+    :param df: pd.DataFrame of the historical data.
+    :param horizon: int number of time into the future for forecasting.
+    :param unit: frequency unit of the time series, which must be a pandas offset alias.
+    :param split_cutoff: the user-specified cutoff, as the starting point of cutoffs
```
wenfeiy-db: maybe make it clear that this is the cutoff between training data and val data.

runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -187,6 +187,33 @@ def generate_cutoffs(df: pd.DataFrame, horizon: int, unit: str,
...
+
+    cutoff = split_cutoff
+    result = [cutoff]
+    while result[-1] <= max(df["ds"]) - horizon_dateoffset:
+        cutoff += period_dateoffset
+        if not (((df["ds"] > cutoff) & (df["ds"] <= cutoff + horizon_dateoffset)).any()):
+            if cutoff < df["ds"].max():
+                closest_date = df[df["ds"] > cutoff].min()["ds"]
+                cutoff = closest_date - horizon_dateoffset
+        result.append(cutoff)
```
wenfeiy-db: Can you add comments like those in L173-182? Would be very helpful for review and when we look back in the future. E.g. is the `cutoff` in `result` guaranteed to be an existing data point in `df` or not?

wenfeiy-db: Also i'm not completely sure whether simply reverting -= to += is just correct. I'll take a deeper look later

runtime/tests/automl_runtime/forecast/pmdarima/training_test.py
```diff
@@ -72,6 +72,17 @@ def test_fit_success_with_exogenous(self):
         results_pd = arima_estimator.fit(self.df_with_exogenous)
         self.assertIn("smape", results_pd)
         self.assertIn("pickled_model", results_pd)
+    
+    def test_fit_success_with_split_cutoff(self):
```
wenfeiy-db, line 76: can we add coverage for different freq like in `test_fit_success`?

runtime/tests/automl_runtime/forecast/prophet/forecast_test.py
```diff
@@ -140,6 +140,33 @@ def test_training_with_extra_regressors(self):
         model_json = json.loads(results["model_json"][0])
         self.assertListEqual(model_json["extra_regressors"][0], ["f1", "f2"])
 
+    def test_training_with_split_cutoff(self):
```
wenfeiy-db, line 143: Can we also add coverage for monthly freq using `self.df_string_monthly_time`?

runtime/tests/automl_runtime/forecast/utils_test.py
```diff
@@ -177,6 +178,72 @@ def test_generate_cutoffs_success_annualy(self):
...
+        self.assertEqual([pd.Timestamp('2012-07-14 00:00:00'), pd.Timestamp('2013-07-14 00:00:00'), pd.Timestamp('2014-07-14 00:00:00')], cutoffs)
+
+    def test_generate_custom_cutoffs_success_with_small_gaps(self):
+        df = pd.DataFrame(
+            pd.date_range(start="2020-07-01", periods=30, freq='3d'), columns=["ds"]
+        ).rename_axis("y").reset_index()
+        cutoffs = generate_custom_cutoffs(df, horizon=7, unit="D", split_cutoff=pd.Timestamp('2020-09-17 00:00:00'))
+        self.assertEqual([pd.Timestamp('2020-09-17 00:00:00'),
+                          pd.Timestamp('2020-09-18 00:00:00'),
+                          pd.Timestamp('2020-09-19 00:00:00')], cutoffs)
```
wenfeiy-db, line 243: Looks like all these three cutoffs actually result to the same training data points because there's no data point of pd.Timestamp('2020-09-18 00:00:00') or pd.Timestamp('2020-09-19 00:00:00'). In this case we probably want to add check so that we don't repeatedly do evaluation using the same data

runtime/databricks/automl_runtime/forecast/pmdarima/training.py
```diff
@@ -35,7 +35,8 @@ class ArimaEstimator:
     """
 
     def __init__(self, horizon: int, frequency_unit: str, metric: str, seasonal_periods: List[int],
-                 num_folds: int = 20, max_steps: int = 150, exogenous_cols: Optional[List[str]] = None) -> None:
+                 num_folds: int = 20, max_steps: int = 150, exogenous_cols: Optional[List[str]] = None,
```
apeforest, line 38: Since you are at it, might as well change the style to 

```
exogenous_cols: List[str] | None = None
```

See https://peps.python.org/pep-0604/

apeforest:
Can you add more details in the PR description

runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -187,6 +187,33 @@ def generate_cutoffs(df: pd.DataFrame, horizon: int, unit: str,
...
+    period = 1 
+    period_dateoffset = pd.DateOffset(**DATE_OFFSET_KEYWORD_MAP[unit])*period
+    horizon_dateoffset = pd.DateOffset(**DATE_OFFSET_KEYWORD_MAP[unit])*horizon
+
+    cutoff = split_cutoff
+    result = [cutoff]
+    while result[-1] <= max(df["ds"]) - horizon_dateoffset:
+        cutoff += period_dateoffset
+        if not (((df["ds"] > cutoff) & (df["ds"] <= cutoff + horizon_dateoffset)).any()):
+            if cutoff < df["ds"].max():
```
apeforest: Combine nested if into one if there is no other branch

runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -187,6 +187,33 @@ def generate_cutoffs(df: pd.DataFrame, horizon: int, unit: str,
...
+    :param split_cutoff: the user-specified cutoff, as the starting point of cutoffs
+    :return: list of pd.Timestamp cutoffs for cross-validation.
+    """
+    period = 1 
+    period_dateoffset = pd.DateOffset(**DATE_OFFSET_KEYWORD_MAP[unit])*period
+    horizon_dateoffset = pd.DateOffset(**DATE_OFFSET_KEYWORD_MAP[unit])*horizon
+
+    cutoff = split_cutoff
+    result = [cutoff]
+    while result[-1] <= max(df["ds"]) - horizon_dateoffset:
```
apeforest: nit: if `df` and `horizon_dataoffset` are not updated inside the while loop, maybe move it before the loop to be more efficient.

runtime/tests/automl_runtime/forecast/utils_test.py
Lanz-db: Do you mean that for each pair of cutoff value, they must have at least one different timestamp in their horizons? I implemented this way because the original generate_cutoff was implemented like this. (no check on evaluating on the same data points) I followed its logic.

I agree with you that we should do this to avoid evaluating on the same training data points multiple times. A ticket has been filed as follow up. ticket: ML-43527

runtime/databricks/automl_runtime/forecast/utils.py
Lanz-db: Fixed.

runtime/databricks/automl_runtime/forecast/utils.py
Lanz-db: Fixed.

runtime/databricks/automl_runtime/forecast/pmdarima/training.py
Lanz-db: Fixed.

runtime/tests/automl_runtime/forecast/prophet/forecast_test.py
Lanz-db: Added.

runtime/tests/automl_runtime/forecast/pmdarima/training_test.py
Lanz-db: Added

runtime/databricks/automl_runtime/forecast/utils.py
Lanz-db: Comments added.

runtime/databricks/automl_runtime/forecast/utils.py
Lanz-db: Added.

runtime/databricks/automl_runtime/forecast/prophet/forecast.py
Lanz-db: Added

runtime/databricks/automl_runtime/forecast/pmdarima/training.py
Lanz-db: Added.

runtime/databricks/automl_runtime/forecast/utils.py
Lanz-db: Filed ML-43528

runtime/databricks/automl_runtime/forecast/pmdarima/training.py
Lanz-db: It seems the python version of this repo does not support the new style.
image
revert it back to Optional

runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -187,6 +187,38 @@ def generate_cutoffs(df: pd.DataFrame, horizon: int, unit: str,
...
+    period_dateoffset = pd.DateOffset(**DATE_OFFSET_KEYWORD_MAP[unit])*period
+    horizon_dateoffset = pd.DateOffset(**DATE_OFFSET_KEYWORD_MAP[unit])*horizon
+
+    # First cutoff is the cutoff bewteen splits
+    cutoff = split_cutoff
+    result = []
+    max_cutoff = max(df["ds"]) - horizon_dateoffset
+    while cutoff <= max_cutoff:
+        # If data does not exist in data range (cutoff, cutoff + horizon_dateoffset]
+        if (not (((df["ds"] > cutoff) & (df["ds"] <= cutoff + horizon_dateoffset)).any())) and (cutoff < df["ds"].max()):
```
wenfeiy-db: Can we remove this redundant condition as we discussed `(cutoff < df["ds"].max()`?

wenfeiy-db (approved):

PR #147, opened 2024-09-27
Remove self._target_col from required columns of DeepAR predict()

es94129:


sabhya-db (approved):
Thx for the quick fix

PR #152, opened 2024-10-08
Fill time steps for time-series in DeepAR predict

es94129:
Currently the `predict` function of DeepAR wouldn't work for datasets loaded from MLflow properly.

- Added `set_index_and_fill_missing_time_steps` utility function, this is used in `predict`, and we shall replace the entire data preprocessing logic before DeepAR training with this too.
- Support single series, multi series (single id_col / multiple id_cols) for `predict`
- Bump version for the next release

Tests
- [x] Pass existing unit tests
- [x] Add unit tests for missing time steps

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/deepar/model.py
+++ b/runtime/databricks/automl_runtime/forecast/deepar/model.py
@@ -124,10 +124,7 @@ class DeepARModel(ForecastModel):
                                                                         self._frequency,
                                                                         self._id_cols)
 
-        if self._id_cols:
-            test_ds = PandasDataset(model_input_transformed, target=self._target_col)
-        else:
-            test_ds = PandasDataset(model_input_transformed, target=self._target_col)
+        test_ds = PandasDataset(model_input_transformed, target=self._target_col)
 
         forecast_iter = self._model.predict(test_ds, num_samples=num_samples)
         forecast_sample_list = list(forecast_iter)
--- a/runtime/databricks/automl_runtime/forecast/deepar/model.py
+++ b/runtime/databricks/automl_runtime/forecast/deepar/model.py
@@ -98,7 +98,8 @@ class DeepARModel(ForecastModel):
 
         pred_df = pred_df.rename(columns={'index': self._time_col})
         if self._id_cols:
-            pred_df = pred_df.rename(columns={'item_id': self._id_cols[0]})
+            id_col_name = '-'.join(self._id_cols)
+            pred_df = pred_df.rename(columns={'item_id': id_col_name})
         else:
             pred_df = pred_df.drop(columns='item_id')
 
--- a/runtime/databricks/automl_runtime/forecast/deepar/model.py
+++ b/runtime/databricks/automl_runtime/forecast/deepar/model.py
@@ -57,10 +57,6 @@ class DeepARModel(ForecastModel):
         :param id_cols: the column names of the identity columns for multi-series time series; None for single series
         """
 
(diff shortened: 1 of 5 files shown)


es94129:
Need to fix the unit tests. They passed in local because my local environment installed pandas 2.x rather than 1.5.3 that AutoML uses.

Edit: Fixed now

runtime/databricks/automl_runtime/forecast/deepar/model.py
```diff
@@ -119,12 +119,15 @@ def predict_samples(self,
         if num_samples is None:
             num_samples = self._num_samples
 
-        model_input = model_input.set_index(self._time_col)
+        model_input_transformed = set_index_and_fill_missing_time_steps(model_input,
+                                                                        self._time_col,
+                                                                        self._frequency,
+                                                                        self._id_cols)
+
         if self._id_cols:
```
sabhya-db: The if and else cases here have the exact same line it seems?

runtime/databricks/automl_runtime/forecast/deepar/utils.py
```diff
@@ -0,0 +1,57 @@
...
+            else:
+                ts_id = str(grouped_id)
+            df_dict[ts_id] = (grouped_df.set_index(time_col).sort_index()
+                              .reindex(new_index_full).drop(id_cols, axis=1))
+
+        return df_dict
+
+    df = df.set_index(time_col).sort_index()
+
+    # Fill in missing time steps between the min and max time steps
```
sabhya-db, line 54: Do we know what the fill_value strategy for the target_col is when we do this?

runtime/tests/automl_runtime/forecast/deepar/utils_test.py
```diff
@@ -0,0 +1,78 @@
...
+        expected_df = base_df.copy()
+        expected_df.loc[[4, 5], target_col] = float('nan')
+        expected_df = expected_df.set_index(time_col).rename_axis(None).asfreq("D")
+
+        pd.testing.assert_frame_equal(transformed_df, expected_df)
+
+    def test_multi_series_filled(self):
+        target_col = "sales"
+        time_col = "date"
+        id_col = "store"
```
sabhya-db, line 53: @es94129 sorry i don't mean to pile onto your work, but can we have a test with multiple id cols too?

sabhya-db (approved):
Thanks @es94129 for the changes, this should simplify jinja templates a lot! Left some minor comments but ow lgtm 🎉

runtime/databricks/automl_runtime/forecast/deepar/model.py
es94129: Good catch, let me simplify it a bit.

runtime/databricks/automl_runtime/forecast/deepar/utils.py
es94129: It creates NaN for the target column. GluonTS will impute the NaN rows for us: https://sourcegraph.com/github.com/awslabs/gluonts/-/blob/src/gluonts/torch/model/deepar/estimator.py?L223

runtime/tests/automl_runtime/forecast/deepar/utils_test.py
es94129: Sure, will do! I also have an end-to-end test in `model_test` with multiple id cols.

sabhya-db: Perfect thank you!

es94129:
Merging the PR with the added test.

PR #154, opened 2024-11-14
Fix weekly date range bug for deepar

es94129:
## What is changed?
Specify the "W-{weekday}" instead of "W" for the `pd.date_range` frequency, so that there is no mismatch between the new index and the original time column's dates.

- Previously: `new_index_full` can start from `2010-02-07` (a Sunday) despite `total_min` is `2010-02-05` (a Friday), because `pd.date_range` assumes that a weekly frequency starts from a Sunday (`W-SUN`)
- Now: the start of `new_index_full` is aligned with `total_min`

## How is it tested?
- [x] Added unit test: provide a dataframe where the dates are Fridays, and the frequency is "W". Before the fix, the unit test would fail with the target column of the output dataframe(s) being all `nan`.
- [x] Manual testing: tested with the Walmart Sales dataset, the DeepAR metrics used to be `NaN` before the change
image


sabhya-db (approved):
lgtm

PR #155, opened 2024-11-15
Take average to aggregate duplicate time_col for DeepAR

es94129:
## What is changed?
Before the change, calling `loaded_model.predict(df)` for a `DeepARModel` would fail if `df` has multiple rows that have the same `time_col` and `id_cols`:

```py
ValueError: cannot reindex on an axis with duplicate labels
```

We add this preprocessing logic inside `predict()` so that users can directly call this function without additional preprocessing to aggregate the dataframe.

## How is it tested?
- [x] Added unit test: without the `groupby` line, this new test will fail with the exact same `ValueError`
- [x] Existing tests work the same
- [x] Manual testing: by changing the `databricks-automl-runtime` version to my uploaded wheel, DeepAR tuning trial notebook completes without errors for `forecasting_tv_sales_small_dataset` without specifying identity columns [job run link](https://e2-dogfood.staging.cloud.databricks.com/jobs/1118598333581389/runs/1?o=6051921418418893), [notebook](https://e2-dogfood.staging.cloud.databricks.com/editor/notebooks/3830593516199824?o=6051921418418893)

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/deepar/model.py
+++ b/runtime/databricks/automl_runtime/forecast/deepar/model.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-import logging
 from typing import List, Optional
 
 import gluonts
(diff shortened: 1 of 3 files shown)


wenfeiy-db (approved):
LGTM. Thanks for the fix!

Just to confirm my understanding—this wasn’t an issue for ARIMA or Prophet because they can still predict without throwing error even when there are two rows with the same time value, right?

es94129:
@wenfeiy-db Yes! This was an edge case that was not tested with DeepAR.

PR #160, opened 2025-01-31
[ML-49316] Support MonthMid and MonthEnd for DeepAR 

Lanz-db:
This PR fixes the bug that when user dataset has monthly frequency and the day of the month is not the first day, DeepAR will fail. The bug results from this line, 
```
new_index_full = pd.date_range(total_min, total_max, freq=frequency)
``` 
`freq` is "MS" so the generated new_index_full will always be the first day of month. So this line, 
```
df.reindex(new_index_full)
``` 
will generate a df with all rows in target column to be NaN.

To fix the bug, this PR introduces
1. a helper function, `validate_and_generate_index `, to generate a complete time index for the given DataFrame based on the specified frequency. If it is monthly frequency, it will generate the index based on the given day of month, also detect if it is the end of month.
2. Use this helper function instead of `pd.date_range(total_min, total_max, freq=frequency)`

To test the function,

- [x] Unit tests added in utils_test.py
Run the below command locally
```
PYTHONPATH=~/automl/runtime/ pytest tests/automl_runtime/forecast/deepar/utils_test.py
```
- [x] manual e2e test
Month start: [notebook](https://e2-dogfood.staging.cloud.databricks.com/editor/notebooks/3625753296542474?o=6051921418418893#command/3625753296542485)
Month mid: [notebook](https://e2-dogfood.staging.cloud.databricks.com/editor/notebooks/430688974618924?o=6051921418418893#command/430688974618931)
Month end: [notebook](https://e2-dogfood.staging.cloud.databricks.com/editor/notebooks/430688974618948?o=6051921418418893#command/4542258616671777)

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/deepar/utils.py
+++ b/runtime/databricks/automl_runtime/forecast/deepar/utils.py
@@ -103,4 +103,10 @@ def set_index_and_fill_missing_time_steps(df: pd.DataFrame, time_col: str,
     df = df.set_index(time_col).sort_index()
 
     # Fill in missing time steps between the min and max time steps
-    return df.reindex(new_index_full)
+    df = df.reindex(new_index_full)
+
+    if frequency.upper() == "MS":
+        # Truncate the day of month to avoid issues with pandas frequency check
+        df = df.to_period("M")
+
+    return df
--- a/runtime/databricks/automl_runtime/forecast/deepar/utils.py
+++ b/runtime/databricks/automl_runtime/forecast/deepar/utils.py
@@ -18,6 +18,50 @@ from typing import List, Optional
 import pandas as pd
 
 
+def validate_and_generate_index(df: pd.DataFrame, time_col: str, frequency: str):
+    """
+    Generate a complete time index for the given DataFrame based on the specified frequency.
+    - Ensures the time column is in datetime format.
+    - Validates consistency in the day of the month if frequency is "MS" (month start).
+    - Generates a new time index from the minimum to the maximum timestamp in the data.
+    :param df: The input DataFrame containing the time column.
+    :param time_col: The name of the time column.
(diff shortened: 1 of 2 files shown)


runtime/tests/automl_runtime/forecast/deepar/utils_test.py
```diff
@@ -143,3 +143,111 @@ def test_single_series_week_day_index(self):
...
+        # Assert equality
+        pd.testing.assert_frame_equal(transformed_df, expected_df)
+
+    def test_single_series_month_end_index(self):
+        target_col = "sales"
+        time_col = "date"
+        num_months = 24
+
+        # Starting from end day of January 2020
+        base_dates = pd.date_range(
```
maggiewang-db, line 226: How is it starting on the last ay of Jan 2020?

runtime/databricks/automl_runtime/forecast/deepar/utils.py
```diff
@@ -18,6 +18,50 @@
 import pandas as pd
 
 
+def validate_and_generate_index(df: pd.DataFrame, time_col: str, frequency: str):
+    """
+    Generate a complete time index for the given DataFrame based on the specified frequency.
```
maggiewang-db, line 23: Thanks for the detailed function description!

runtime/tests/automl_runtime/forecast/deepar/utils_test.py
Lanz-db: see line 228, by specifying freq='M', it is by default the end of the month

maggiewang-db: I see. Can you add a comment on the line below?

maggiewang-db (approved):
Thank you for fixing it

PR #162, opened 2025-02-11
[ML-47076] Support custom frequencies

Lanz-db:
This PR added support to custom frequencies in AutoML. It added `frequency_quantity` to the Estimator and Model class for all three algorithms. To review this PR, please review it with the PR, https://github.com/databricks-eng/universe/pull/911199.

To test this PR

- [x] unit tests
- [x] manual e2e tests

To manually test, I

1. installed core library wheel and automl_runtime wheel in a MLR 14.3 LTS compute
2. commented out the line %pip install --force-reinstall databricks-automl-runtime==0.2.20.5 in pip_install.jinja and pip_install_deepar.jinja to use the new automl_runtime wheel
3. ran the driver notebooks
I tested all possible combinations, stored under the [folder](https://e2-dogfood.staging.cloud.databricks.com/browse/folders/3282835429133688?o=6051921418418893), including
5 min - single time series
5 min - multi time series
10 min - single time series
10 min - multi time series
15 min - single time series
15 min - multi time series
30 min - single time series
30 min - multi time series


runtime/databricks/automl_runtime/forecast/deepar/utils.py
```diff
@@ -18,7 +18,7 @@
 import pandas as pd
 
 
-def validate_and_generate_index(df: pd.DataFrame, time_col: str, frequency: str):
+def validate_and_generate_index(df: pd.DataFrame, time_col: str, frequency: str, frequency_quantity: int):
```
apeforest: please also update docstring

runtime/databricks/automl_runtime/forecast/deepar/utils.py
```diff
@@ -86,7 +87,7 @@ def set_index_and_fill_missing_time_steps(df: pd.DataFrame, time_col: str,
         weekday_name = total_min.strftime("%a").upper() # e.g., "FRI"
         frequency = f"W-{weekday_name}"
 
-    new_index_full = validate_and_generate_index(df=df, time_col=time_col, frequency=frequency)
+    new_index_full = validate_and_generate_index(df=df, time_col=time_col, frequency=frequency, frequency_quantity=frequency_quantity)
```
apeforest: nit: break into multilines and indent

apeforest: nit: naming is not clear. What does new_index_full mean? especially the "_full"  suffix.

runtime/databricks/automl_runtime/forecast/deepar/utils.py
apeforest: Why do we need to add a new parameter to this function? Can we process the frequecy as f"{frequency_quantity}{frequency}" and pass into this function directly?

The internal `pd.date_range` takes a freq as string so it doesn't really need the frequency_quantity as input from this function.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -64,7 +64,7 @@ def model_env(self):
         return ARIMA_CONDA_ENV
 
     @staticmethod
-    def _get_ds_indices(start_ds: pd.Timestamp, periods: int, frequency: str) -> pd.DatetimeIndex:
+    def _get_ds_indices(start_ds: pd.Timestamp, periods: int, frequency: str, frequency_quantity: int) -> pd.DatetimeIndex:
```
apeforest: Same as above. frequency_quantity is not needed.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -268,7 +270,7 @@ class MultiSeriesArimaModel(AbstractArimaModel):
     ARIMA mlflow model wrapper for multivariate forecasting.
     """
 
-    def __init__(self, pickled_model_dict: Dict[Tuple, bytes], horizon: int, frequency: str,
+    def __init__(self, pickled_model_dict: Dict[Tuple, bytes], horizon: int, frequency: str, frequency_quantity: int,
```
apeforest: Same here. ArimaModel classs can directly take in frequency string

apeforest (changes requested):
Can we reduce the number of changes to the function interface, but calculate frequency string from the caller?

runtime/databricks/automl_runtime/forecast/deepar/utils.py
Lanz-db: It is true that some functions only need one frequency string by inputing f"{frequency_quantity}{frequency}". However, there are some functions that need a separate integer frequency_quantity to do some math calculation. So my point was to make the interface consistent, I do not want frequency some time to be only min some time to be 5min, which will make the code hard to read I believe.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: Same concern above

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: Same concern above

runtime/databricks/automl_runtime/forecast/deepar/utils.py
Lanz-db: updated

runtime/databricks/automl_runtime/forecast/deepar/utils.py
Lanz-db: This PR does not change the name. But I agree that it is not clear. Rename it to valid_index.

Lanz-db: Done

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
apeforest: Here it makes sense to have a separate field.

runtime/databricks/automl_runtime/forecast/deepar/model.py
```diff
@@ -42,7 +42,7 @@ class DeepARModel(ForecastModel):
     DeepAR mlflow model wrapper for forecasting.
     """
 
-    def __init__(self, model: PyTorchPredictor, horizon: int, frequency: str,
+    def __init__(self, model: PyTorchPredictor, horizon: int, frequency: str, frequency_quantity: int,
```
apeforest: I see in the training.py, the `frequency` field is called `frequency_unit` which makes better sense. Let's use that across this change.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
apeforest: Ack

runtime/databricks/automl_runtime/forecast/deepar/utils.py
```diff
@@ -18,7 +18,10 @@
 import pandas as pd
 
 
-def validate_and_generate_index(df: pd.DataFrame, time_col: str, frequency: str):
+def validate_and_generate_index(df: pd.DataFrame, 
+                                time_col: str,
```
apeforest, line 22: I understand you need to pass in frequency unit and frequency quantity at high-level (e.g. ForecastModel) as the config. However, it's not necessary to keep them down to this util level.

runtime/databricks/automl_runtime/forecast/pmdarima/training.py
```diff
@@ -36,7 +36,7 @@ class ArimaEstimator:
 
     def __init__(self, horizon: int, frequency_unit: str, metric: str, seasonal_periods: List[int],
                  num_folds: int = 20, max_steps: int = 150, exogenous_cols: Optional[List[str]] = None,
-                 split_cutoff: Optional[pd.Timestamp] = None) -> None:
+                 split_cutoff: Optional[pd.Timestamp] = None, frequency_quantity: int = 1) -> None:
```
apeforest, line 39: Can we put `frequency_quantity` closer to `frequency_unit` for better readability.

runtime/databricks/automl_runtime/forecast/prophet/forecast.py
```diff
@@ -41,7 +41,9 @@ def _prophet_fit_predict(params: Dict[str, Any], history_pd: pd.DataFrame,
                          horizon: int, frequency: str, cutoffs: List[pd.Timestamp],
                          interval_width: int, primary_metric: str,
                          country_holidays: Optional[str] = None,
-                         regressors = None, **prophet_kwargs) -> Dict[str, Any]:
+                         regressors = None, 
+                         frequency_quantity: int = 1,
```
apeforest, line 45: put frequency_quantity closer to frequency_unit (renamed)

runtime/databricks/automl_runtime/forecast/pmdarima/training.py
Lanz-db: I intentionally added it in the end because I want frequency_quantity to have a default value = 1 to make it backward compatible. With default value, it cannot be placed before any argument without default value. Even if it is the first one among arguments with default value, it is still far from frequency_unit so I put it in the end.

runtime/databricks/automl_runtime/forecast/prophet/forecast.py
Lanz-db: I intentionally added it in the end because I want frequency_quantity to have a default value = 1 to make it backward compatible. With default value, it cannot be placed before any argument without default value. Even if it is the first one among arguments with default value, it is still far from frequency_unit so I put it in the end.

runtime/tests/automl_runtime/forecast/deepar/utils_test.py
```diff
@@ -39,7 +39,7 @@ def test_single_series_filled(self):
         )
         dropped_df = base_df.drop([4, 5]).reset_index(drop=True)
 
-        transformed_df = set_index_and_fill_missing_time_steps(dropped_df, time_col, "D")
+        transformed_df = set_index_and_fill_missing_time_steps(dropped_df, time_col, "D", 1)
```
apeforest, line 43: Please test different quantities

runtime/tests/automl_runtime/forecast/deepar/model_test.py
```diff
@@ -94,6 +94,7 @@ def test_model_save_and_load_single_series(self):
             model=self.model,
             horizon=self.prediction_length,
             frequency="d",
+            frequency_quantity=1,
```
apeforest, line 98: Please test different quantities

runtime/databricks/automl_runtime/forecast/deepar/utils.py
apeforest: This is non-blocking, but only hope to reduce the amont of code change in this PR.

apeforest (commented):
Please rename frequency to frequency_unit and add more tests

Lanz-db:
> Please rename frequency to frequency_unit and add more tests

Thanks for the review. To clarify, I am working on adding tests, the commits I pushed was to fix the existing tests. Will let you know once I added tests.

runtime/tests/automl_runtime/forecast/deepar/model_test.py
Lanz-db: tests added.

runtime/tests/automl_runtime/forecast/deepar/utils_test.py
Lanz-db: tests added

runtime/databricks/automl_runtime/forecast/deepar/model.py
Lanz-db: Renamed all frequency in this repo to frequency_unit.

runtime/databricks/automl_runtime/forecast/deepar/utils.py
Lanz-db: I would prefer separating them to make the interface consistent.

runtime/databricks/automl_runtime/forecast/prophet/forecast.py
apeforest: Can we instead pass in Frequency as a dataclass?

```
@dataclass
class Frequency:
    """Class to specify forecast frequency."""
    frequency_unit: str
    frequency_value: int = 1
```

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -45,21 +45,23 @@ class ProphetModel(ForecastModel):
     Prophet mlflow model wrapper for univariate forecasting.
     """
 
-    def __init__(self, model_json: Union[Dict[Tuple, str], str], horizon: int, frequency: str,
+    def __init__(self, model_json: Union[Dict[Tuple, str], str], horizon: int, frequency_unit: str, frequency_quantity: int,
```
apeforest: nit: please make one argument per line.

runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -74,45 +77,46 @@ def make_single_future_dataframe(
...
         start_time = start_time
     else:
-        start_time = end_time + unit_offset
+        start_time = end_time + timestep_offset
 
     date_rng = pd.date_range(
         start=start_time,
-        end=end_time + unit_offset*horizon,
-        freq=unit_offset
+        end=end_time + timestep_offset*horizon,
```
apeforest: nit: space around * operator

runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -133,36 +137,38 @@ def get_validation_horizon(df: pd.DataFrame, horizon: int, unit: str, frequency_
...
-    if unit in NON_DAILY_OFFSET_ALIAS:
+    if frequency_unit in NON_DAILY_OFFSET_ALIAS:
         period = int(period)
-        period_dateoffset = pd.DateOffset(**DATE_OFFSET_KEYWORD_MAP[unit])*period
+        period_dateoffset = pd.DateOffset(**DATE_OFFSET_KEYWORD_MAP[frequency_unit])*frequency_quantity*period
     else:
-        offset_kwarg = {list(DATE_OFFSET_KEYWORD_MAP[unit])[0]: period}
-        period_dateoffset = pd.DateOffset(**offset_kwarg)
+        offset_kwarg = {list(DATE_OFFSET_KEYWORD_MAP[frequency_unit])[0]: period}
+        period_dateoffset = pd.DateOffset(**offset_kwarg)*frequency_quantity
```
apeforest: nit: space around * operator

runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -239,30 +247,33 @@ def is_frequency_consistency(
...
+    :param frequency_quantity: the multiplier for the frequency.
     :return: A boolean indicate whether the time interval is
              evenly divisible by the period.
     """
-    periods = calculate_period_differences(start_time, end_time, freq)
+    periods = calculate_period_differences(start_time, end_time, frequency_unit, frequency_quantity)
     diff = pd.to_datetime(end_time) -  pd.DateOffset(
-                **DATE_OFFSET_KEYWORD_MAP[OFFSET_ALIAS_MAP[freq]]
-            ) * periods == pd.to_datetime(start_time)
+                **DATE_OFFSET_KEYWORD_MAP[OFFSET_ALIAS_MAP[frequency_unit]]
```
apeforest, line 257: Can we clean up the logic or add comment here. It's hard to understand what this does

runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -239,30 +247,33 @@ def is_frequency_consistency(
...
+    :param frequency_quantity: An integer that is the multiplier for the frequency.
     :return: A pd.Series indicates the round-down integer period
              calculated.
     """
     start_time = pd.to_datetime(start_time)
     end_time = pd.to_datetime(end_time)
-    freq_alias = PERIOD_ALIAS_MAP[OFFSET_ALIAS_MAP[freq]]
-    return  (end_time.to_period(freq_alias) - start_time.to_period(freq_alias)).n
+    freq_alias = PERIOD_ALIAS_MAP[OFFSET_ALIAS_MAP[frequency_unit]]
+    return  (end_time.to_period(freq_alias) - start_time.to_period(freq_alias)).n // frequency_quantity
```
apeforest, line 281: what if it's not divisible, we just get the floor value? is that intended?

runtime/tests/automl_runtime/forecast/pmdarima/model_test.py
```diff
@@ -460,7 +472,8 @@ def test_mlflow_arima_log_model_multiseries(self):
         end_ds_dict = {("1",): pd.Timestamp("2020-10-09"), ("2",): pd.Timestamp("2020-10-09")}
         multiseries_arima_model = MultiSeriesArimaModel(pickled_model_dict,
                                                         horizon=1,
-                                                        frequency='d',
+                                                        frequency_unit='d',
+                                                        frequency_quantity=1,
```
apeforest, line 476: indent?

runtime/tests/automl_runtime/forecast/pmdarima/training_test.py
```diff
@@ -48,13 +48,37 @@ def setUp(self) -> None:
...
+            ['d', 1, self.df, [1, 7]], 
+            ['d', 1, self.df_string_time, [1, 7]], 
+            ['month', 1, self.df_monthly, [1, 12]],
+            ['min', 5, self.df_with_5_minute_interval, [1]],
+            ['min', 10, self.df_with_10_minute_interval, [1]],
+            ['min', 15, self.df_with_15_minute_interval, [1]],
+            ['min', 30, self.df_with_30_minute_interval, [1]]]:
             arima_estimator = ArimaEstimator(horizon=1,
                                              frequency_unit=freq,
+                                             frequency_quantity=frequancy_quantity,
```
apeforest, line 79: nit: indent?

runtime/tests/automl_runtime/forecast/pmdarima/training_test.py
```diff
@@ -74,13 +99,18 @@ def test_fit_success_with_exogenous(self):
...
+                         ['min', 5, self.df_with_5_minute_interval, '2020-07-05 00:30:00'],
+                         ['min', 10, self.df_with_10_minute_interval, '2020-07-05 01:00:00'],
+                         ['min', 15, self.df_with_15_minute_interval, '2020-07-05 01:30:00'],
+                         ['min', 30, self.df_with_30_minute_interval, '2020-07-05 03:00:00']]:
             arima_estimator = ArimaEstimator(horizon=1,
                                             frequency_unit=freq,
+                                            frequency_quantity=frequency_quantity,
                                             metric="smape",
-                                            seasonal_periods=[1, 7],
+                                            seasonal_periods=[1],
```
apeforest: why change seasonal_periods?

apeforest (commented):
I suggest to use a dataclass to represent frequency_unit and frequency_value as one entity instead of treating them as separate arguments.

nit: please run lint on your code. There are many indentation issue and multiple arguments in one line.

runtime/databricks/automl_runtime/forecast/utils.py
Lanz-db: yes it is intended. If it is not divisible, we will get the floor value. And in the later check we will use this floor value to find out it is not consistent.

runtime/databricks/automl_runtime/forecast/prophet/forecast.py
Lanz-db: ok the problem is that for some functions, I want frequency_quantity has default value = 1. These functions are the ones that are used outside, in the trial notebook. If a customer developed based on an old trial notebook, we do not want to break those function calls in the old trial notebooks. 

For other functions that are only called inside this repo, I recommend having no default values so it can expose bugs earlier. Sometimes with default value, some bugs may be silently ignored.

Besides, I believe we are around the corner to refactor this repo and core library code, as we now are pretty sure we will go black-box. So I would suggest deferring introducing any complex data structures for now.

runtime/databricks/automl_runtime/forecast/prophet/model.py
Lanz-db: Done

runtime/databricks/automl_runtime/forecast/utils.py
Lanz-db: Done

runtime/databricks/automl_runtime/forecast/utils.py
Lanz-db: Done

runtime/databricks/automl_runtime/forecast/utils.py
Lanz-db: Comments added

Lanz-db: it is to calculate if the difference between start and end time is divisible by the period time

runtime/tests/automl_runtime/forecast/pmdarima/model_test.py
Lanz-db: I do not quite understand here. These arguments are aligned by their first letter, right?

runtime/tests/automl_runtime/forecast/pmdarima/training_test.py
Lanz-db: Same, these arguments are aligned by their first letter.

runtime/tests/automl_runtime/forecast/pmdarima/training_test.py
Lanz-db: typo. Fixed.

runtime/databricks/automl_runtime/forecast/prophet/forecast.py
apeforest: Is this code shared between serverless and classic? If so, then it makes sense. If not, why worried about trial notebook?

Lanz-db: Oh I was wrong, you are right. This code is only used by serverless.
Let me clarify the tradeoffs of introducing a new dataclass.
Pros: code is cleaner. (But I doubt how much it improved the code readability. I believe without the dataclass, it is still pretty readable)
Cons: 
1. Need to change a lot of code and their related tests, including some codes that are not necessarily touched in this PR.
2. This dataclass will force me to either all have default values or all without default values. But we actually do not need to have default values now. But anyway it makes the code lose flexibility.

I would still prefer not using such dataclass. wdyt?

apeforest: It's not just cleaner code (in terms of readability) but better for overall maintainability and testability.

Think about the following:
1. The current argument list is already long. For the consumer of this function, they need to read many argument definitions in order to understand how to use this
2. The frequency_unit and frequency_quantity is a combined logic concept. Specify either one of the doesn't make sense.
3. Easy to validate: you can add a validate function inside the dataclass. So every time a Frequency object is passed in, it can be validated immediately.
4. Easy test: you only need to test the Frequency object instead of two separate values. You can have a simple test like `def is_valid_frequency(freq: Frequency) -> bool`. In the future, if you want to support other quantities, e.g. 45min or 3d, you don't need to add check all over the place.
5. Having a default value in it is not a bad thing. With single object, you can change the default value only once and it applies to all.

Anyway, given the urgency of this task, I don't want to make this a blocker. But I suggest you create a follow-up JIRA in EE epic to refactor this code.

runtime/databricks/automl_runtime/forecast/utils.py
apeforest: please add comment in here.

runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -239,30 +247,34 @@ def is_frequency_consistency(
...
              evenly divisible by the period.
     """
-    periods = calculate_period_differences(start_time, end_time, freq)
+    periods = calculate_period_differences(start_time, end_time, frequency_unit, frequency_quantity)
+    # If the difference between start and end time is divisible by the period time
     diff = pd.to_datetime(end_time) -  pd.DateOffset(
-                **DATE_OFFSET_KEYWORD_MAP[OFFSET_ALIAS_MAP[freq]]
-            ) * periods == pd.to_datetime(start_time)
+                **DATE_OFFSET_KEYWORD_MAP[OFFSET_ALIAS_MAP[frequency_unit]]
+            ) * periods * frequency_quantity == pd.to_datetime(start_time)
```
apeforest: Still not clear to me. What is this `==` check doing here? Please also add parenthesis around to make the compute precedence clearer.

apeforest (approved):
Overall, looks good. But the code still needs refactoring. Please add comments to two places I mentioned and create a JIRA for follow-up refactoring.

runtime/databricks/automl_runtime/forecast/prophet/forecast.py
Lanz-db: jira ticket, https://databricks.atlassian.net/jira/software/c/projects/ML/boards/3737?selectedIssue=ML-50316

runtime/databricks/automl_runtime/forecast/utils.py
Lanz-db: on the left of this ==, it calculates the difference between end time and period time, on the right, it is start time. the variable diff is True or False

PR #169, opened 2025-04-25
[ML-52574] Add covariate support in predict_timeseries for prediction table usage

Lanz-db:
This PR adds covariate support in predict_timeseries function for prediction table usage.
In details, it
1. adds preprocess_func before doing prediction for covariates
2. fix the reset_index bug that id column can not be inserted back due to duplicate

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/prophet/model.py
+++ b/runtime/databricks/automl_runtime/forecast/prophet/model.py
@@ -156,6 +156,30 @@ class ProphetModel(ForecastModel):
         test_df.rename(columns={self._time_col: "ds"}, inplace=True)
         predict_df = self.model().predict(test_df)
         return predict_df["yhat"]
+    
+    def predict(self, model_input: pd.DataFrame) -> pd.DataFrame:
+        """
+        Predict API for prediction tables with covariates
+        :param model_input: Input dataframe
+        :return: A pd.DataFrame with the forecast components.
+        """
+        self._validate_cols(model_input, [self._time_col])
+        test_df = model_input.copy()
+
+        if self._preprocess_func and self._split_col:
+            # Apply the same preprocessing pipeline to test_df. The preprocessing function requires the "y" column 
+            # and the split column to be present, as they are used in the trial notebook. These columns are added 
+            # temporarily and removed after preprocessing.
+            # see https://src.dev.databricks.com/databricks-eng/universe/-/blob/automl/python/databricks/automl/core/sections/templates/preprocess/finish_with_transform.jinja?L3
+            # and https://src.dev.databricks.com/databricks-eng/universe/-/blob/automl/python/databricks/automl/core/sections/templates/preprocess/select_columns.jinja?L8-10
+            test_df["y"] = None
+            test_df[self._split_col] = "prediction"
+            test_df = self._preprocess_func(test_df)
+            test_df.drop(columns=["y", self._split_col], inplace=True, errors="ignore")
+
+        test_df.rename(columns={self._time_col: "ds"}, inplace=True)
+        predict_df = self.model().predict(test_df)
+        return predict_df
 
     def infer_signature(self, sample_input: pd.DataFrame = None) -> ModelSignature:
         if sample_input is None:
@@ -353,6 +377,43 @@ class MultiSeriesProphetModel(ProphetModel):
         predict_df = test_df.groupby(self._id_cols).apply(model_prediction).reset_index(drop=True)
         return_df = test_df.merge(predict_df, how="left", on=["ds"] + self._id_cols)
         return return_df["yhat"]
+    
+    def predict(self, model_input: pd.DataFrame) -> pd.DataFrame:
+        """
+        Predict API for prediction tables with covariates
+        :param model_input: Input dataframe
+        :return: A pd.DataFrame with the forecast components.
+        """
+        self._validate_cols(model_input, self._id_cols + [self._time_col])
+        test_df = model_input.copy()
+        test_df["ts_id"] = test_df[self._id_cols].apply(tuple, axis=1)
+
+        if self._preprocess_func and self._split_col:
+            # Apply the same preprocessing pipeline to test_df. The preprocessing function requires the "y" column 
+            # and the split column to be present, as they are used in the trial notebook. These columns are added 
+            # temporarily and removed after preprocessing.
+            # see https://src.dev.databricks.com/databricks-eng/universe/-/blob/automl/python/databricks/automl/core/sections/templates/preprocess/finish_with_transform.jinja?L3
+            # and https://src.dev.databricks.com/databricks-eng/universe/-/blob/automl/python/databricks/automl/core/sections/templates/preprocess/select_columns.jinja?L8-10
+            test_df["y"] = None
+            test_df[self._split_col] = ""
+            test_df = test_df.groupby(self._id_cols).apply(self._preprocess_func).reset_index(drop=True)
+            test_df.drop(columns=["y", self._split_col], inplace=True, errors="ignore")
+
+        test_df.rename(columns={self._time_col: "ds"}, inplace=True)
+
+        def model_prediction(df):
+            model = self.model(df["ts_id"].iloc[0])
+            if model:
+                predicts = model.predict(df)
+                # We have to explicitly assign the id columns to avoid KeyError when model_input
+                # only has one row. For multi-rows model_input, the ts_id will be kept as index
+                # after groupby(self._id_cols).apply(...) and we can retrieve it by reset_index, but
+                # for one-row model_input the id columns are missing from index.
+                predicts[self._id_cols] = df.name
+                return predicts
+        predict_df = test_df.groupby(self._id_cols).apply(model_prediction).reset_index(drop=True)
+        return_df = test_df.merge(predict_df, how="left", on=["ds"] + self._id_cols)
+        return return_df
 
 
 def mlflow_prophet_log_model(prophet_model: Union[ProphetModel, MultiSeriesProphetModel],
--- a/runtime/databricks/automl_runtime/forecast/prophet/model.py
+++ b/runtime/databricks/automl_runtime/forecast/prophet/model.py
@@ -110,7 +110,7 @@ class ProphetModel(ForecastModel):
                                                   freq=pd.DateOffset(**offset_kwarg),
                                                   include_history=include_history)
 
-    def _predict_impl(self, horizon: int = None, include_history: bool = True, df: pd.DataFrame | None = None) -> pd.DataFrame:
+    def _predict_impl(self, horizon: int = None, include_history: bool = True, df: pd.DataFrame = None) -> pd.DataFrame:
         """
         Predict using the API from prophet model.
         :param horizon: Int number of periods to forecast forward.
@@ -129,7 +129,7 @@ class ProphetModel(ForecastModel):
         future_pd.rename(columns={self._time_col: "ds"}, inplace=True)
         return self.model().predict(future_pd)
 
-    def predict_timeseries(self, horizon: int = None, include_history: bool = True, df: pd.DataFrame | None = None) -> pd.DataFrame:
+    def predict_timeseries(self, horizon: int = None, include_history: bool = True, df: pd.DataFrame = None) -> pd.DataFrame:
         """
         Predict using the prophet model.
         :param horizon: Int number of periods to forecast forward.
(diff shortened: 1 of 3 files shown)


runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -156,6 +156,30 @@ def predict(self, context: mlflow.pyfunc.model.PythonModelContext, model_input:
...
+            # see https://src.dev.databricks.com/databricks-eng/universe/-/blob/automl/python/databricks/automl/core/sections/templates/preprocess/finish_with_transform.jinja?L3
+            # and https://src.dev.databricks.com/databricks-eng/universe/-/blob/automl/python/databricks/automl/core/sections/templates/preprocess/select_columns.jinja?L8-10
+            test_df["y"] = None
+            test_df[self._split_col] = "prediction"
+            test_df = self._preprocess_func(test_df)
+            test_df.drop(columns=["y", self._split_col], inplace=True, errors="ignore")
+
+        test_df.rename(columns={self._time_col: "ds"}, inplace=True)
+        predict_df = self.model().predict(test_df)
+        return predict_df
```
apeforest: Is this the only difference from `def predict` above? 

1. why do we need other field? Can we concatenate them with predict_df['yhat'] outside of the call?
2. this function looks almost identical from the one above. can we refactor if answer to (1) is justified.

apeforest (changes requested):
Please address the comment before merging.

runtime/databricks/automl_runtime/forecast/prophet/model.py
Lanz-db: 1. We need other field such as xxx_upper and xxx_lower for the predicted values, which was a previous customer request. The old predict function does not return upper and lower bounds. We cannot concatenate outside of the call, this is the only source we can get other fields.
2. We cannot refactor the old predict function, as the old function comment says, it is to override mlflow model's predict function, we shall not change the return type.

Lanz-db: Oh what we can do is to call the new function predict_with_full_df_returned in the old function. Does it sound good to you?

apeforest: But in your notebook (in the other universe PR), it also shows xxx_upper and xxx_lower for feature columns.

Lanz-db: > it also shows xxx_upper and xxx_lower for feature columns.

Yes, it is expected, they are the output of the model.

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -110,26 +110,34 @@ def make_future_dataframe(self, horizon: int = None, include_history: bool = Tru
                                                   freq=pd.DateOffset(**offset_kwarg),
                                                   include_history=include_history)
 
-    def _predict_impl(self, horizon: int = None, include_history: bool = True) -> pd.DataFrame:
+    def _predict_impl(self, horizon: int = None, include_history: bool = True, df: pd.DataFrame = None) -> pd.DataFrame:
```
apeforest: nit: rename df to future_df or future_feature_df

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -110,26 +110,34 @@ def make_future_dataframe(self, horizon: int = None, include_history: bool = Tru
...
         """
         Predict using the API from prophet model.
         :param horizon: Int number of periods to forecast forward.
         :param include_history: Boolean to include the historical dates in the data
             frame for predictions.
         :return: A pd.DataFrame with the forecast components.
```
apeforest: please update docstring

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -110,26 +110,34 @@ def make_future_dataframe(self, horizon: int = None, include_history: bool = Tru
...
+            future_pd = self.make_future_dataframe(horizon=horizon or self._horizon, include_history=include_history)
+
+        if self._preprocess_func and self._split_col:
+            future_pd = apply_preprocess_func(future_pd, self._preprocess_func, self._split_col)
+
+        future_pd.rename(columns={self._time_col: "ds"}, inplace=True)
         return self.model().predict(future_pd)
 
-    def predict_timeseries(self, horizon: int = None, include_history: bool = True) -> pd.DataFrame:
+    def predict_timeseries(self, horizon: int = None, include_history: bool = True, df: pd.DataFrame = None) -> pd.DataFrame:
```
apeforest: rename df to future_df and update docstring.

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -260,28 +260,35 @@ def _predict_impl(self, df: pd.DataFrame, horizon: int = None, include_history:
         future_pd[self._id_cols] = df[self._id_cols].iloc[0]
         return future_pd
 
-    def predict_timeseries(self, horizon: int = None, include_history: bool = True) -> pd.DataFrame:
+    def predict_timeseries(self, horizon: int = None, include_history: bool = True, df: pd.DataFrame = None) -> pd.DataFrame:
```
apeforest: rename df to future_df and update docstring

runtime/databricks/automl_runtime/forecast/prophet/model.py
Lanz-db: updated

runtime/databricks/automl_runtime/forecast/prophet/model.py
Lanz-db: updated

runtime/databricks/automl_runtime/forecast/prophet/model.py
Lanz-db: updated

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -110,26 +110,36 @@ def make_future_dataframe(self, horizon: int = None, include_history: bool = Tru
...
                                                   include_history=include_history)
 
-    def _predict_impl(self, horizon: int = None, include_history: bool = True) -> pd.DataFrame:
+    def _predict_impl(self, horizon: int = None, include_history: bool = True, future_df: pd.DataFrame = None) -> pd.DataFrame:
         """
         Predict using the API from prophet model.
         :param horizon: Int number of periods to forecast forward.
         :param include_history: Boolean to include the historical dates in the data
             frame for predictions.
+        :param future_df: Optional future input dataframe
```
apeforest: be more explicit that the future_df is for covariates and contains the covariate columns.

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -110,26 +110,36 @@ def make_future_dataframe(self, horizon: int = None, include_history: bool = Tru
...
         :return: A pd.DataFrame with the forecast components.
         """
-        future_pd = self.make_future_dataframe(horizon=horizon or self._horizon, include_history=include_history)
+        if future_df is not None:
```
apeforest: nit: refactor
```
if future_pd is None:
  future_pd = self.make_future_dataframe(horizon=horizon or self._horizon, include_history=include_history)
```

runtime/databricks/automl_runtime/forecast/prophet/model.py
Lanz-db: actually future_df can be without covariates. The implementation here enables the function to take a df as input, it does not matter if it contains covariates

apeforest: ic. please add more comments to the doc string: future_df is the dataframe that contains the time series column and covaraite columns if available. The returned result will contain the copy of future_df and the predicted target column.

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -110,26 +110,34 @@ def make_future_dataframe(self, horizon: int = None, include_history: bool = Tru
...
         :return: A pd.DataFrame with the forecast components.
         """
-        future_pd = self.make_future_dataframe(horizon=horizon or self._horizon, include_history=include_history)
-        return self.model().predict(future_pd)
+        if future_df is None:
+            future_df = self.make_future_dataframe(horizon=horizon or self._horizon, include_history=include_history)
 
-    def predict_timeseries(self, horizon: int = None, include_history: bool = True) -> pd.DataFrame:
+        if self._preprocess_func and self._split_col:
+            future_df = apply_preprocess_func(future_df, self._preprocess_func, self._split_col)
```
apeforest, line 140: When I read the function, this is a side effect that takes in the class field.

I suggest passing `preprocess_fun` and `split_col` as arguments to the _predict_impl and moving this logic to `predict_timeseries` so they are more consistent between prophet and arima.

Also, please document in the `predict_timeseries` that the preprocess will be applied.

runtime/databricks/automl_runtime/forecast/utils.py
```diff
@@ -279,3 +279,21 @@ def calculate_period_differences(
     freq_alias = PERIOD_ALIAS_MAP[OFFSET_ALIAS_MAP[frequency_unit]]
     # It is intended to get the floor value. And in the later check we will use this floor value to find out if it is not consistent.
     return  (end_time.to_period(freq_alias) - start_time.to_period(freq_alias)).n // frequency_quantity
+
+def apply_preprocess_func(df: pd.DataFrame, preprocess_func: callable, split_col: str) -> pd.DataFrame:
```
apeforest, line 283: Non-blocking in this PR. it's always strange to read that split_col is required for such function. Let's refactor it in a follow-up

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -110,26 +110,34 @@ def make_future_dataframe(self, horizon: int = None, include_history: bool = Tru
...
+
+        future_df.rename(columns={self._time_col: "ds"}, inplace=True)
+        return self.model().predict(future_df)
+
+    def predict_timeseries(self, horizon: int = None, include_history: bool = True, future_df: pd.DataFrame = None) -> pd.DataFrame:
```
apeforest: please add unit test to test the future_df is used or created if None

runtime/databricks/automl_runtime/forecast/prophet/model.py
```diff
@@ -260,28 +260,33 @@ def _predict_impl(self, df: pd.DataFrame, horizon: int = None, include_history:
...
+                frequency_quantity=self._frequency_quantity,
+                include_history=include_history,
+                groups=self._model_json.keys(),
+                identity_column_names=self._id_cols
+            )
         future_df["ts_id"] = future_df[self._id_cols].apply(tuple, axis=1)
-        return future_df.groupby(self._id_cols).apply(lambda df: self._predict_impl(df, horizon, include_history)).reset_index()
+        if self._preprocess_func and self._split_col:
+            future_df = apply_preprocess_func(future_df, self._preprocess_func, self._split_col)
+        future_df.rename(columns={self._time_col: "ds"}, inplace=True)
```
apeforest, line 293: unit test the future_df has columns renanmed

apeforest (commented):
please refactor the predict_impl and add unit test.

runtime/databricks/automl_runtime/forecast/prophet/model.py
Lanz-db: preprocess_func and split_col are already class variables, why do we want to make them as input argument to another class function?

Just to clarify, this PR did not break consistency more between prophet and arima. Previously, arima has its make_futuer_dataframe inside _predict_impl but not predict_timeseries, which is not consistent with prophet.

runtime/databricks/automl_runtime/forecast/prophet/model.py
Lanz-db: added

runtime/databricks/automl_runtime/forecast/prophet/model.py
Lanz-db: moved

runtime/databricks/automl_runtime/forecast/prophet/model.py
Lanz-db: added

runtime/databricks/automl_runtime/forecast/prophet/model.py
Lanz-db: added

runtime/tests/automl_runtime/forecast/prophet/model_test.py
```diff
@@ -451,3 +488,68 @@ def preprocess_func(df):
         )
         yhat = prophet_model.predict(None, test_df)
         self.assertEqual(2, len(yhat))
+
+    @patch("databricks.automl_runtime.forecast.prophet.model.MultiSeriesProphetModel._predict_impl")
```
apeforest, line 492: nit: indentation of @patch line should be aligned with def test... below?

apeforest (approved):
Thanks for the changes.

runtime/tests/automl_runtime/forecast/prophet/model_test.py
Lanz-db: it is aligned on my side. Or python will complain?

PR #170, opened 2025-04-26
[ML-52570] Add covariate inference support in automl runtime for ARIMA

Lanz-db:
This PR modifies model.py under ARIMA to support inference with covariates. In details, it
1. adds the usage preprocess_func in predict
2. modifies the function argument for predict_timeseries to align with Prophet

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/pmdarima/model.py
+++ b/runtime/databricks/automl_runtime/forecast/pmdarima/model.py
@@ -150,13 +150,13 @@ class ArimaModel(AbstractArimaModel):
         if self._preprocess_func and self._split_col:
             future_df = apply_preprocess_func(future_df, self._preprocess_func, self._split_col)
         horizon = horizon or self._horizon
-        X = None
+        future_feature_df = None
         if self._exogenous_cols and future_df is not None:
             time_col = self._time_col if self._time_col in future_df.columns else "ds"
-            X = (future_df[future_df[time_col] > self._end_ds].set_index(time_col))[self._exogenous_cols]
-        future_pd = self._forecast(horizon, X)
+            future_feature_df = (future_df[future_df[time_col] > self._end_ds].set_index(time_col))[self._exogenous_cols]
+        future_pd = self._forecast(horizon, future_feature_df)
         if include_history:
-            in_sample_pd = self._predict_in_sample(start_ds=self._start_ds, end_ds=self._end_ds, X=X)
+            in_sample_pd = self._predict_in_sample(start_ds=self._start_ds, end_ds=self._end_ds, feature_df=future_feature_df)
             return pd.concat([in_sample_pd, future_pd]).reset_index(drop=True)
         else:
             return future_pd
@@ -226,18 +226,18 @@ class ArimaModel(AbstractArimaModel):
         # Out-of-sample prediction if needed
         horizon = calculate_period_differences(self._end_ds, max(df["ds"]), self._frequency_unit, self._frequency_quantity)
         if horizon > 0:
-            X_future = df[df["ds"] > self._end_ds].set_index("ds")
+            future_feature_df = df[df["ds"] > self._end_ds].set_index("ds")
             future_pd = self._forecast(
                 horizon,
-                X=X_future[self._exogenous_cols] if self._exogenous_cols else None)
+                feature_df=future_feature_df[self._exogenous_cols] if self._exogenous_cols else None)
             preds_pds.append(future_pd)
         # In-sample prediction if needed
         if pred_start_ds <= self._end_ds:
-            X_in_sample = df[df["ds"] <= self._end_ds].set_index("ds")
+            df_in_sample = df[df["ds"] <= self._end_ds].set_index("ds")
             in_sample_pd = self._predict_in_sample(
                 start_ds=pred_start_ds,
                 end_ds=self._end_ds,
-                X=X_in_sample[self._exogenous_cols] if self._exogenous_cols else None)
+                feature_df=df_in_sample[self._exogenous_cols] if self._exogenous_cols else None)
             preds_pds.append(in_sample_pd)
         # Map predictions back to given timestamps
         preds_pd = pd.concat(preds_pds).set_index("ds")
@@ -248,7 +248,7 @@ class ArimaModel(AbstractArimaModel):
         self,
         start_ds: pd.Timestamp = None,
         end_ds: pd.Timestamp = None,
-        X: pd.DataFrame = None) -> pd.DataFrame:
+        feature_df: pd.DataFrame = None) -> pd.DataFrame:
         if start_ds and end_ds:
             start_idx = calculate_period_differences(self._start_ds, start_ds, self._frequency_unit, self._frequency_quantity)
             end_idx = calculate_period_differences(self._start_ds, end_ds, self._frequency_unit, self._frequency_quantity)
@@ -259,7 +259,7 @@ class ArimaModel(AbstractArimaModel):
         d = self.model().order[1]
         start_idx = max(start_idx, d)
         preds_in_sample, conf_in_sample = self.model().predict_in_sample(
-            X=X,
+            X=feature_df,
             start=start_idx,
             end=end_idx,
             return_conf_int=True)
@@ -272,12 +272,12 @@ class ArimaModel(AbstractArimaModel):
     def _forecast(
         self,
         horizon: int = None,
-        X: pd.DataFrame = None) -> pd.DataFrame:
+        feature_df: pd.DataFrame = None) -> pd.DataFrame:
         # set horizon to the length of future_df if future_df is provided to avoid the length mismatch error from pmdarima
-        horizon = horizon or self._horizon if X is None else len(X)
+        horizon = horizon or self._horizon if feature_df is None else len(feature_df)
         preds, conf = self.model().predict(
             horizon,
-            X=X,
+            X=feature_df,
             return_conf_int=True)
         ds_indices = self._get_ds_indices(start_ds=self._end_ds, periods=horizon + 1, frequency_unit=self._frequency_unit, frequency_quantity=self._frequency_quantity)[1:]
         preds_pd = pd.DataFrame({'ds': ds_indices, 'yhat': preds})
--- a/runtime/databricks/automl_runtime/forecast/pmdarima/model.py
+++ b/runtime/databricks/automl_runtime/forecast/pmdarima/model.py
@@ -273,6 +273,7 @@ class ArimaModel(AbstractArimaModel):
         self,
         horizon: int = None,
         feature_df: pd.DataFrame = None) -> pd.DataFrame:
+        # Unlike Prophet, pmdarima does not require time column in the feature_df, it depends on horizon to determine the length of the prediction
         # set horizon to the length of future_df if future_df is provided to avoid the length mismatch error from pmdarima
         horizon = horizon or self._horizon if feature_df is None else len(feature_df)
         preds, conf = self.model().predict(
--- a/runtime/databricks/automl_runtime/forecast/pmdarima/model.py
+++ b/runtime/databricks/automl_runtime/forecast/pmdarima/model.py
@@ -90,9 +90,17 @@ class ArimaModel(AbstractArimaModel):
     ARIMA mlflow model wrapper for univariate forecasting.
     """
 
-    def __init__(self, pickled_model: bytes, horizon: int, frequency_unit: str,
-                 frequency_quantity: int, start_ds: pd.Timestamp, end_ds: pd.Timestamp,
-                 time_col: str, exogenous_cols: Optional[List[str]] = None) -> None:
+    def __init__(self, 
+                 pickled_model: bytes, 
+                 horizon: int, 
+                 frequency_unit: str,
+                 frequency_quantity: int, 
+                 start_ds: pd.Timestamp, 
+                 end_ds: pd.Timestamp,
+                 time_col: str, 
+                 exogenous_cols: Optional[List[str]] = None,
+                 split_col: Optional[str] = None,
+                 preprocess_func: Optional[callable] = None) -> None:
         """
         Initialize the mlflow Python model wrapper for ARIMA.
         :param pickled_model: the pickled ARIMA model as a bytes object.
@@ -114,6 +122,8 @@ class ArimaModel(AbstractArimaModel):
         self._end_ds = pd.to_datetime(end_ds)
         self._time_col = time_col
         self._exogenous_cols = exogenous_cols
+        self._split_col = split_col
+        self._preprocess_func = preprocess_func
 
     def model(self) -> pmdarima.arima.ARIMA:
         """
@@ -178,8 +188,44 @@ class ArimaModel(AbstractArimaModel):
         :return: A pd.Series with the prediction values.
         """
         self._validate_cols(model_input, [self._time_col])
-        result_df = self._predict_impl(model_input)
+        test_df = model_input.copy()
+        if self._preprocess_func and self._split_col:
+             # Apply the same preprocessing pipeline to test_df. The preprocessing function requires the "y" column 
+            # and the split column to be present, as they are used in the trial notebook. These columns are added 
+            # temporarily and removed after preprocessing.
+            # see https://src.dev.databricks.com/databricks-eng/universe/-/blob/automl/python/databricks/automl/core/sections/templates/preprocess/finish_with_transform.jinja?L3
+            # and https://src.dev.databricks.com/databricks-eng/universe/-/blob/automl/python/databricks/automl/core/sections/templates/preprocess/select_columns.jinja?L8-10
+            test_df["y"] = None
+            test_df[self._split_col] = "prediction"
+            test_df = self._preprocess_func(test_df)
+            test_df.drop(columns=["y", self._split_col], inplace=True, errors="ignore")
+        result_df = self._predict_impl(test_df)
         return result_df["yhat"]
+
+    def predict_with_full_df_returned(self, model_input: pd.DataFrame) -> pd.DataFrame:
+        """
+        Predict API for prediction tables with covariates.
+
+        Returns the prediction values for given timestamps in the input dataframe. If an input timestamp
+        to predict does not match the original frequency that the model trained on, an exception will be thrown.
+        :param model_input: The input dataframe of the model. Should have the same time column name
+                            as the training data of the ARIMA model.
+        :return: A pd.DataFrame with the prediction values.
+        """
+        self._validate_cols(model_input, [self._time_col])
+        test_df = model_input.copy()
+        if self._preprocess_func and self._split_col:
+             # Apply the same preprocessing pipeline to test_df. The preprocessing function requires the "y" column 
+            # and the split column to be present, as they are used in the trial notebook. These columns are added 
+            # temporarily and removed after preprocessing.
+            # see https://src.dev.databricks.com/databricks-eng/universe/-/blob/automl/python/databricks/automl/core/sections/templates/preprocess/finish_with_transform.jinja?L3
+            # and https://src.dev.databricks.com/databricks-eng/universe/-/blob/automl/python/databricks/automl/core/sections/templates/preprocess/select_columns.jinja?L8-10
+            test_df["y"] = None
+            test_df[self._split_col] = "prediction"
+            test_df = self._preprocess_func(test_df)
+            test_df.drop(columns=["y", self._split_col], inplace=True, errors="ignore")
+        result_df = self._predict_impl(test_df)
+        return result_df
 
     def _predict_impl(self, input_df: pd.DataFrame) -> pd.DataFrame:
         df = input_df.rename(columns={self._time_col: "ds"})
@@ -272,9 +318,18 @@ class MultiSeriesArimaModel(AbstractArimaModel):
     ARIMA mlflow model wrapper for multivariate forecasting.
     """
 
-    def __init__(self, pickled_model_dict: Dict[Tuple, bytes], horizon: int, frequency_unit: str, frequency_quantity: int,
-                 start_ds_dict: Dict[Tuple, pd.Timestamp], end_ds_dict: Dict[Tuple, pd.Timestamp],
-                 time_col: str, id_cols: List[str], exogenous_cols: Optional[List[str]] = None) -> None:
+    def __init__(self, 
+                 pickled_model_dict: Dict[Tuple, bytes], 
+                 horizon: int, 
+                 frequency_unit: str, 
+                 frequency_quantity: int,
+                 start_ds_dict: Dict[Tuple, pd.Timestamp], 
+                 end_ds_dict: Dict[Tuple, pd.Timestamp],
+                 time_col: str, 
+                 id_cols: List[str], 
+                 exogenous_cols: Optional[List[str]] = None,
+                 split_col: Optional[str] = None,
+                 preprocess_func: Optional[callable] = None) -> None:
         """
         Initialize the mlflow Python model wrapper for multiseries ARIMA.
         :param pickled_model_dict: the dictionary of binarized ARIMA models for different time series.
@@ -298,6 +353,8 @@ class MultiSeriesArimaModel(AbstractArimaModel):
         self._time_col = time_col
         self._id_cols = id_cols
         self._exogenous_cols = exogenous_cols
+        self._split_col = split_col
+        self._preprocess_func = preprocess_func
 
     def model(self, id_: Tuple) -> pmdarima.arima.ARIMA:
         """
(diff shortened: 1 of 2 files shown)


runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -126,20 +138,22 @@ def predict_timeseries(
...
         Predict target column for given horizon_timedelta and history data.
         :param horizon: int number of periods to forecast forward.
         :param include_history: Boolean to include the historical dates in the data
             frame for predictions.
-        :param df: A pd.Dataframe containing regressors (exogenous variables), if they were used to train the model.
+        :param future_df: A pd.Dataframe containing regressors (exogenous variables), if they were used to train the model.
         :return: A pd.DataFrame with the forecasts and confidence intervals for given horizon_timedelta and history data.
         """
+        if self._preprocess_func and self._split_col:
+            future_df = apply_preprocess_func(future_df, self._preprocess_func, self._split_col)
```
apeforest, line 152: what is future_df is None?

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -126,20 +138,22 @@ def predict_timeseries(
...
         horizon = horizon or self._horizon
         X = None
-        if self._exogenous_cols and df is not None:
-            time_col = self._time_col if self._time_col in df.columns else "ds"
-            X = (df[df[time_col] > self._end_ds].set_index(time_col))[self._exogenous_cols]
+        if self._exogenous_cols and future_df is not None:
```
apeforest: if future_df already contains the covariates, what _exogenous_cols are for here?

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -90,9 +90,17 @@ class ArimaModel(AbstractArimaModel):
...
-                 time_col: str, exogenous_cols: Optional[List[str]] = None) -> None:
+    def __init__(self, 
+                 pickled_model: bytes, 
+                 horizon: int, 
+                 frequency_unit: str,
+                 frequency_quantity: int, 
+                 start_ds: pd.Timestamp, 
+                 end_ds: pd.Timestamp,
+                 time_col: str, 
+                 exogenous_cols: Optional[List[str]] = None,
```
apeforest, line 101: It seems my previous review was not stored here...

Can we remove such argument and do the same way as prophet? I don't think we ever used this argument. Can we add the covariates in the data frame in core library.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -126,20 +138,22 @@ def predict_timeseries(
...
+        if self._preprocess_func and self._split_col:
+            future_df = apply_preprocess_func(future_df, self._preprocess_func, self._split_col)
         horizon = horizon or self._horizon
         X = None
-        if self._exogenous_cols and df is not None:
-            time_col = self._time_col if self._time_col in df.columns else "ds"
-            X = (df[df[time_col] > self._end_ds].set_index(time_col))[self._exogenous_cols]
+        if self._exogenous_cols and future_df is not None:
+            time_col = self._time_col if self._time_col in future_df.columns else "ds"
+            X = (future_df[future_df[time_col] > self._end_ds].set_index(time_col))[self._exogenous_cols]
```
apeforest: If we remove the _exogenous_cols, we don't need this `X` variable here,

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -272,9 +290,18 @@ class MultiSeriesArimaModel(AbstractArimaModel):
...
+    def __init__(self, 
+                 pickled_model_dict: Dict[Tuple, bytes], 
+                 horizon: int, 
+                 frequency_unit: str, 
+                 frequency_quantity: int,
+                 start_ds_dict: Dict[Tuple, pd.Timestamp], 
+                 end_ds_dict: Dict[Tuple, pd.Timestamp],
+                 time_col: str, 
+                 id_cols: List[str], 
+                 exogenous_cols: Optional[List[str]] = None,
```
apeforest, line 341: same for multiseries

apeforest (commented):
Let's follow the same pattern as prophet.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: Discussed offline. The current implementation can reuse the most logics between Prophet and ARIMA, the slight difference here (exogenous_cols) is already the best way.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: When future_df is None, self._preprocess_func and self._split_col will be None.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: Discussed offline. _exogenous_cols is used to select covariates from future_df, required by pmdarima model

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: changed X to other name

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
apeforest: Where is this predicate validated?

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
apeforest: Please add a TODO here. If Arima can accept future_df without dropping the time column, then the code can be simplified a lot here.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: preprocess_func, split_col, and future_df are only set when there is covariate. It is the logic in the jinja in core lib

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: added

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -114,6 +124,8 @@ def __init__(self, pickled_model: bytes, horizon: int, frequency_unit: str,
         self._end_ds = pd.to_datetime(end_ds)
         self._time_col = time_col
         self._exogenous_cols = exogenous_cols
+        self._split_col = split_col
```
apeforest, line 127: Add a #TODO comment here to remove split_col. We should handle it outside of the runtime. The current argument seems very confusing.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -209,18 +226,18 @@ def _predict_impl(self, input_df: pd.DataFrame) -> pd.DataFrame:
         # Out-of-sample prediction if needed
         horizon = calculate_period_differences(self._end_ds, max(df["ds"]), self._frequency_unit, self._frequency_quantity)
         if horizon > 0:
-            X_future = df[df["ds"] > self._end_ds].set_index("ds")
+            future_feature_df = df[df["ds"] > self._end_ds].set_index("ds")
             future_pd = self._forecast(
                 horizon,
-                X=X_future[self._exogenous_cols] if self._exogenous_cols else None)
+                feature_df=future_feature_df[self._exogenous_cols] if self._exogenous_cols else None)
```
apeforest, line 261: nit: add space around =

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: actually I do not think we can get rid of split_col. As long as preprocess_func requires split_col, we will need split_col because we will have to apply preprocess_func.

Besides, I do not quite get it why it is confusing. split_col is like time_col, it is a property of the training data. I see them the same.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -255,11 +272,13 @@ def _predict_in_sample(
     def _forecast(
         self,
         horizon: int = None,
-        X: pd.DataFrame = None) -> pd.DataFrame:
-        horizon = horizon or self._horizon
+        feature_df: pd.DataFrame = None) -> pd.DataFrame:
+        # Unlike Prophet, pmdarima does not require time column in the feature_df, it depends on horizon to determine the length of the prediction
+        # set horizon to the length of future_df if future_df is provided to avoid the length mismatch error from pmdarima
+        horizon = horizon or self._horizon if feature_df is None else len(feature_df)
```
apeforest: avoid such complex ternary operator, change it to if conditioin

```
if feature_df is None:
    horizon = horizon or self._horizon
else
    horizon = len(feature_df)
```

apeforest: another question, does `len(feature_df)` always return the number of rows in feature_df?

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
apeforest: Add a #TODO here to remove such argument if Arima can take the entire dataframe

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: yes

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: added in line 154

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: fixed for the entire file.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: Done

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -126,23 +138,25 @@ def predict_timeseries(
...
-            X = (df[df[time_col] > self._end_ds].set_index(time_col))[self._exogenous_cols]
-        future_pd = self._forecast(horizon, X)
+        future_feature_df = None
+        if self._exogenous_cols and future_df is not None:
+            time_col = self._time_col if self._time_col in future_df.columns else "ds"
+            future_feature_df = (future_df[future_df[time_col] > self._end_ds].set_index(time_col))[self._exogenous_cols]
+        future_pd = self._forecast(horizon, future_feature_df)
         if include_history:
-            in_sample_pd = self._predict_in_sample(start_ds=self._start_ds, end_ds=self._end_ds, X=X)
+            in_sample_pd = self._predict_in_sample(start_ds=self._start_ds, end_ds=self._end_ds, feature_df=future_feature_df)
```
apeforest: Can you check if such call is correct? based on the [API](https://alkaline-ml.com/pmdarima/modules/generated/pmdarima.arima.ARIMA.html#pmdarima.arima.ARIMA.predict_in_sample), this is to run prediction on the historical data. Then we don't need he future_feature_df, right?

Lanz-db: True. Should input None.
And actually, for now, we never call predict_timeseries with include_history=True

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
apeforest: Let's add an assert or check. We should specify this in contract not replying on knowledge.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -126,24 +138,27 @@ def predict_timeseries(
...
         include_history: bool = True,
-        df: Optional[pd.DataFrame] = None) -> pd.DataFrame:
+        future_df: Optional[pd.DataFrame] = None) -> pd.DataFrame:
         """
         Predict target column for given horizon_timedelta and history data.
         :param horizon: int number of periods to forecast forward.
         :param include_history: Boolean to include the historical dates in the data
             frame for predictions.
-        :param df: A pd.Dataframe containing regressors (exogenous variables), if they were used to train the model.
+        :param future_df: A pd.Dataframe containing regressors (exogenous variables), if they were used to train the model.
```
apeforest: Change it to: `A pd.Dataframe containing future time indices and future covariates if covariates are used in training.`

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -126,24 +138,27 @@ def predict_timeseries(
         self,
         horizon: int = None,
         include_history: bool = True,
-        df: Optional[pd.DataFrame] = None) -> pd.DataFrame:
+        future_df: Optional[pd.DataFrame] = None) -> pd.DataFrame:
         """
         Predict target column for given horizon_timedelta and history data.
```
apeforest: nit change horizon_timedelta to horizon

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -126,24 +138,27 @@ def predict_timeseries(
...
         :param horizon: int number of periods to forecast forward.
         :param include_history: Boolean to include the historical dates in the data
             frame for predictions.
-        :param df: A pd.Dataframe containing regressors (exogenous variables), if they were used to train the model.
+        :param future_df: A pd.Dataframe containing regressors (exogenous variables), if they were used to train the model.
         :return: A pd.DataFrame with the forecasts and confidence intervals for given horizon_timedelta and history data.
```
apeforest: nit change horizon_timedelta to horizon

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
apeforest: If it's used only for process_func, we should somehow package it with process_fun instead of having two arguments. Let's find a better way to refactor this in the future. Please add TODO

The naming suggests it's how we split the train/validate/test in the dataset. It's not clear why only preprocess needs it.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -126,24 +138,27 @@ def predict_timeseries(
...
         horizon = horizon or self._horizon
-        X = None
-        if self._exogenous_cols and df is not None:
-            time_col = self._time_col if self._time_col in df.columns else "ds"
-            X = (df[df[time_col] > self._end_ds].set_index(time_col))[self._exogenous_cols]
-        future_pd = self._forecast(horizon, X)
+        future_feature_df = None
+        # TODO: investigate if we can use future_df directly in the forecast function
+        if self._exogenous_cols and future_df is not None:
+            time_col = self._time_col if self._time_col in future_df.columns else "ds"
```
apeforest: move this to a separate function: 

def drop_time_column(df: DataFrame) -> DataFrame

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -126,24 +138,27 @@ def predict_timeseries(
...
+            future_feature_df = (future_df[future_df[time_col] > self._end_ds].set_index(time_col))[self._exogenous_cols]
+        future_pd = self._forecast(horizon, future_feature_df)
         if include_history:
-            in_sample_pd = self._predict_in_sample(start_ds=self._start_ds, end_ds=self._end_ds, X=X)
-            return pd.concat([in_sample_pd, future_pd]).reset_index(drop=True)
+            in_sample_pd = self._predict_in_sample(start_ds = self._start_ds, end_ds = self._end_ds, feature_df = None)
```
apeforest: nit: I think the assignment inside function don't need space around?

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -126,24 +138,27 @@ def predict_timeseries(
...
+            return pd.concat([in_sample_pd, future_pd]).reset_index(drop = True)
```
apeforest: nit: remove space around =

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -157,12 +172,12 @@ def make_future_dataframe(self, horizon: int = None, include_history: bool = Tru
         requested number of periods.
         """
         return make_single_future_dataframe(
-            start_time=self._start_ds,
-            end_time=self._end_ds,
-            horizon=horizon or self._horizon,
-            frequency_unit=self._frequency_unit,
-            frequency_quantity=self._frequency_quantity,
-            include_history=include_history
+            start_time = self._start_ds,
```
apeforest: nit: no space within function call

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -178,60 +193,63 @@ def predict(self, context: mlflow.pyfunc.model.PythonModelContext, model_input:
...
+        test_df = model_input.copy()
+        if self._preprocess_func and self._split_col:
+            test_df = apply_preprocess_func(test_df, self._preprocess_func, self._split_col)
+        result_df = self._predict_impl(test_df)
         return result_df["yhat"]
 
     def _predict_impl(self, input_df: pd.DataFrame) -> pd.DataFrame:
-        df = input_df.rename(columns={self._time_col: "ds"})
-        df["ds"] = pd.to_datetime(df["ds"], infer_datetime_format=True)
+        df = input_df.rename(columns = {self._time_col: "ds"})
```
apeforest: same here

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -178,60 +193,63 @@ def predict(self, context: mlflow.pyfunc.model.PythonModelContext, model_input:
...
+        df["ds"] = pd.to_datetime(df["ds"], infer_datetime_format = True)
```
apeforest: same here

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
apeforest: should the feature_df here be the training df without the future features? https://alkaline-ml.com/pmdarima/modules/generated/pmdarima.arima.ARIMA.html#pmdarima.arima.ARIMA.predict_in_sample

apeforest: need to drop time column in feature_df, and pass in the new df here.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: added

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: moved

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: in your previous comment, you suggested to add space around = in function call, https://github.com/databricks/automl/pull/170#discussion_r2070642201

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: done

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: modified

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: changed

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
apeforest: my bad 😢  in the earlier review, I thought it was a single assignment not a keyword arg. 
image

According to  https://peps.python.org/pep-0008/, it's better to not wrap space around kwargs assignment: 
Can you please change it back. Sorry for the confusion.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -126,27 +138,44 @@ def predict_timeseries(
...
-            in_sample_pd = self._predict_in_sample(start_ds=self._start_ds, end_ds=self._end_ds, X=X)
-            return pd.concat([in_sample_pd, future_pd]).reset_index(drop=True)
+            history_feature_df = self._get_history_feature_df(future_df)
+            in_sample_pd = self._predict_in_sample(start_ds = self._start_ds, end_ds = self._end_ds, feature_df = history_feature_df)
+            return pd.concat([in_sample_pd, future_pd]).reset_index(drop = True)
         else:
             return future_pd
 
+    def _get_future_feature_df(self, future_df: pd.DataFrame) -> pd.DataFrame:
+        if future_df is not None:
```
apeforest, line 172: nit: can you add one line docstring: return the features from the data frame without time column?

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -126,27 +138,44 @@ def predict_timeseries(
...
+            time_col = self._time_col if self._time_col in future_df.columns else "ds"
+            return (future_df[future_df[time_col] > self._end_ds].set_index(time_col))[self._exogenous_cols]
+        else:
+            return None
+    
+    def _get_history_feature_df(self, future_df: pd.DataFrame = None) -> pd.DataFrame:
+        if future_df is not None:
+            return future_df[future_df[self._time_col] <= self._end_ds].set_index(self._time_col)[self._exogenous_cols]
```
apeforest: what would happen if future_df[future_df[self._time_col] <= self._end_ds] returns empty? i.e. the future_df doesn't contain any history data.

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -255,13 +287,21 @@ def _predict_in_sample(
     def _forecast(
         self,
         horizon: int = None,
-        X: pd.DataFrame = None) -> pd.DataFrame:
-        horizon = horizon or self._horizon
+        feature_df: pd.DataFrame = None) -> pd.DataFrame:
+        """
+
```
apeforest: please fill in the docstring here?

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
```diff
@@ -332,34 +385,37 @@ def make_future_dataframe(
...
+        :param future_df: A pd.Dataframe containing regressors (exogenous variables), if they were used to train the model.
         :return: A pd.DataFrame with the forecast components.
         """
         horizon = horizon or self._horizon
         ids = self._pickled_models.keys()
-        preds_dfs = list(map(lambda id_: self._predict_timeseries_single_id(id_, horizon, include_history, df), ids))
-        return pd.concat(preds_dfs).reset_index(drop=True)
+        if future_df is not None:
+            # Add ts_id column to future_df because preprocess_func requires it
+            future_df["ts_id"] = future_df[self._id_cols].apply(tuple, axis=1)
```
apeforest, line 433: when will this be dropped?

Lanz-db: will be dropped in core lib

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: filled

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: add assert to prevent this

runtime/databricks/automl_runtime/forecast/pmdarima/model.py
Lanz-db: added

apeforest (approved):
Thank you for the refactoring

PR #176, opened 2025-06-26
[ES-1509817] Incorrect forecast horizon in batch inference using DeepAR

Lanz-db:
Fix a bug in batch inference notebook. The bug is that DeepAR forecast results have incorrect time column. The bug is due to the call of `to_timestamp()` which will by default use the first day of the period, but we should use the last day of the period.

### DIFF
--- a/runtime/databricks/automl_runtime/forecast/deepar/model.py
+++ b/runtime/databricks/automl_runtime/forecast/deepar/model.py
@@ -107,7 +107,7 @@ class DeepARModel(ForecastModel):
         else:
             pred_df = pred_df.drop(columns='item_id')
 
-        pred_df[self._time_col] = pred_df[self._time_col].dt.to_timestamp()
+        pred_df[self._time_col] = pred_df[self._time_col].dt.end_time.dt.normalize()
 
         return pred_df
 
--- a/runtime/databricks/automl_runtime/forecast/deepar/model.py
+++ b/runtime/databricks/automl_runtime/forecast/deepar/model.py
@@ -107,7 +107,10 @@ class DeepARModel(ForecastModel):
         else:
             pred_df = pred_df.drop(columns='item_id')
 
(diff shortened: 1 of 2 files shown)


runtime/databricks/automl_runtime/forecast/deepar/model.py
```diff
@@ -107,7 +107,10 @@ def predict(self,
         else:
             pred_df = pred_df.drop(columns='item_id')
 
-        pred_df[self._time_col] = pred_df[self._time_col].dt.to_timestamp()
+        if self._frequency_unit == 'W':
```
apeforest: Can you move this logic into a function and add comments? E.g.
```
pred_df = self.normalize_time_column(pred_df)
```
And add unit test for this function.

apeforest:
Can you provide more explanation in the description:

> The bug is due to the call of to_timestamp() which will by default use the first day of the period, but we should use the last day of the period.

Why we should use the last day of the period only when frequency is W? Also, why this was not an issue for prophet and arima?

runtime/databricks/automl_runtime/forecast/deepar/model.py
Lanz-db: moved and added

Lanz-db:
> Can you provide more explanation in the description:
> 
> > The bug is due to the call of to_timestamp() which will by default use the first day of the period, but we should use the last day of the period.
> 
> Why we should use the last day of the period only when frequency is W? Also, why this was not an issue for prophet and arima?

because the pred_df returned by deepar is like below. The index (time column) is a period object, where end date is the correct day of the week we expect.
```
                   index      yhat item_id
0  2024-07-03/2024-07-09  0.672345    None
```

apeforest:
Thank you! LGTM

apeforest (approved):

PR #177, opened 2025-09-29
[ML-52572] Add covariate inference support in automl runtime for deepar

Pajaraja:
Updated DeepARModel to include the feature (covariate) columns, and a function to preprocess columns.

Added tests to check new functionality.


apeforest (approved):
LGTM

